PR #20, opened 2017-06-17
Bugfix: Loop bodies

mbeckersys:
Bodies of for-loops had been ignored, and bodies of while-loops caused a crash due to calling the wrong function in an attempt to append the ireps of the loop body.

Updated one test case (while) and added another one (for). Related to this, added an optional parameter to the prove function in the test scripts, to handover cbmc flags like "--unwind".

Cbmc fails to analyze some loops, yet, because Boolean conditions are encoded in an incompatible way (will propose fix for Booleans separately).


mbeckersys:
While being correct, this was not everything. Loop conditions are not translated here. Will open another pull request.

PR #21, opened 2017-07-07
Fix #19 Subtype indication without constraint

### LINKED ISSUE #19: Subtype indication without constraint raises exception
Hi,

The tree walker fails on derived types without subtype constraints, at least in the following specific context:
```
procedure main is
  type myint is new Integer;
  x : myint := 2;
  function inc (x : myint) return myint is (x+1);
begin
  x := inc (x);
end main;
```
produces the error message
```
GNU Ada (ada2goto) Assert_Failure failed precondition from tree_walk.adb:203
Error detected at test.adb:1:11
```
The precondition is that `Nkind (N) = N_Subtype_Indication`, but the actual node kind is `N_Identifier`.

mbeckersys:
GNAT does not produce N_Subtype_Indication when no constraint is given, but provides the identifier directly. This fix checks for the type of the subtype indication node, and acts accordingly. Including test case.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -200,7 +200,7 @@ package body Tree_Walk is
    with Pre => Nkind (N) = N_Subtype_Declaration;
 
    function Do_Subtype_Indication (N : Node_Id) return Irep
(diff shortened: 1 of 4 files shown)


smowton:
Looks great, thanks @mbeckersys !

PR #24, opened 2017-07-24
Feature/nondet

mbeckersys:
Every function with prefix `nondet` is modeled as providing a non-deterministic return value, whereas the value is in the domain of the according type (i.e., respecting its ranges). This also works when such a function is part of an expression. Details here: https://github.com/mbeckersys/gnat2goto/wiki/Non-Determinism

Some other extensions are here, too: detecting calls of RTS functions (we don't have a body for those), as well as allowing null procedures.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -103,9 +103,9 @@ package body Tree_Walk is
    with Pre  => Nkind (N) = N_Function_Call,
         Post => Kind (Do_Nondet_Function_Call'Result) in Class_Expr;
 
-   function Make_Sym_Range_Expression (I : Irep) return Irep with
-     Pre => Kind (I) = I_Symbol_Expr,
-     Post => Kind (Make_Sym_Range_Expression'Result) in Class_Expr;
+   function Make_Sym_Range_Expression (I : Irep) return Irep
+   with Pre  => Kind (I) = I_Symbol_Expr,
+        Post => Kind (Make_Sym_Range_Expression'Result) in Class_Expr;
 
    function Do_Handled_Sequence_Of_Statements (N : Node_Id) return Irep
    with Pre  => Nkind (N) = N_Handled_Sequence_Of_Statements,
@@ -1146,25 +1146,6 @@ package body Tree_Walk is
       end if;
    end Do_Full_Type_Declaration;
 
-   ---------------------
-   -- Name_Has_Prefix --
-   ---------------------
-
-   function Name_Has_Prefix (N : Node_Id; Prefix : String) return Boolean
-     with Pre => Present (Name (N));
-
-   function Name_Has_Prefix (N : Node_Id; Prefix : String) return Boolean is
-      Short_Name : constant String :=
-        Get_Name_String (Chars (Name (N)));
-   begin
-      if Prefix'Length > Short_Name'Length then
-         return False;
-      else
-         return Prefix = Short_Name
-           (Short_Name'First .. Short_Name'First - 1 + Prefix'Length);
-      end if;
-   end Name_Has_Prefix;
-
    -------------------------------
    -- Make_Sym_Range_Expression --
    -------------------------------
@@ -1181,8 +1162,8 @@ package body Tree_Walk is
 
       if Kind (Sym_Type) = I_Bounded_Signedbv_Type then
          declare
-            Op_Geq   : constant Irep := New_Irep (I_Op_Geq);
-            Op_Leq   : constant Irep := New_Irep (I_Op_Leq);
+            Op_Geq : constant Irep := New_Irep (I_Op_Geq);
+            Op_Leq : constant Irep := New_Irep (I_Op_Leq);
          begin
             Set_Lhs (Op_Geq, I);
             Set_Rhs (Op_Geq, Get_Lower_Bound (Sym_Type));
@@ -1212,10 +1193,8 @@ package body Tree_Walk is
    function Do_Nondet_Function_Call (N : Node_Id) return Irep is
       Func_Str     : constant String := Unique_Name (Entity (Name (N)));
       Func_Name    : constant Symbol_Id := Intern (Func_Str);
-      Func_Symbol  : Symbol;
    begin
       if Global_Symbol_Table.Contains (Func_Name) then
-         Func_Symbol := Global_Symbol_Table (Func_Name);
          --  ??? why not get this from the entity
 
          --  Two implementation options here:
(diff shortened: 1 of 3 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -1130,6 +1138,130 @@ package body Tree_Walk is
       end if;
    end Do_Full_Type_Declaration;
 
+   ---------------------
+   -- Name_Has_Prefix --
+   ---------------------
+
+   function Name_Has_Prefix (N : Node_Id; Prefix : String) return Boolean
```
smowton: Kick this into utils I'd say

mbeckersys: I agree, this is again one of those cases where it would be more natural to implement at the cbmc side, especially because the range checks are already inserted there. But also a bit trickier, I suppose. 
That is just a temporary solution (now that I've said this, chances are this will stick around for a while). On the plus side, the symbol table is a bit smaller this way.

The other changes related to null procedures and RTS calls, however, should still be merged.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1130,6 +1138,118 @@ package body Tree_Walk is
...
+     with Pre => Present (Name (N));
+
+   function Name_Has_Prefix (N : Node_Id; Prefix : String) return Boolean is
+      Short_Name : constant String :=
+        Get_Name_String (Chars (Name (N)));
+   begin
+      if Prefix'Length > Short_Name'Length then
```
ptroja: it would be slightly less verbose to write

`return Prefix'Length <= Short_Name'Length and then Prefix = Short_Name (...);`

but it is a matter of style (though in general, Ada is already quite verbose, so it makes sense to write concisely)

gnat2goto/driver/tree_walk.adb
```diff
@@ -1130,6 +1138,118 @@ package body Tree_Walk is
...
+   begin
+      --  get underlying type
+      while Kind (Sym_Type) = I_Symbol_Type loop
+         Sym_Type := Follow_Symbol_Type
+           (Sym_Type, Global_Symbol_Table);
+      end loop;
+
+      if Kind (Sym_Type) = I_Bounded_Signedbv_Type then
+         declare
+            Op_Geq   : constant Irep := New_Irep (I_Op_Geq);
```
ptroja: please remove extra spaces before colons

gnat2goto/driver/tree_walk.adb
```diff
@@ -1130,6 +1138,118 @@ package body Tree_Walk is
...
+   -- Do_Nondet_Function_Call --
+   -----------------------------
+
+   function Do_Nondet_Function_Call (N : Node_Id) return Irep is
+      Func_Str     : constant String := Unique_Name (Entity (Name (N)));
+      Func_Name    : constant Symbol_Id := Intern (Func_Str);
+      Func_Symbol  : Symbol;
+   begin
+      if Global_Symbol_Table.Contains (Func_Name) then
+         Func_Symbol := Global_Symbol_Table (Func_Name);
```
ptroja: this could be constant in the `declare` block below (?)

smowton:
@mbeckersys you seen @ptroja's comments above?

mbeckersys:
Yes, I have seen this, but forgot to answer. Thanks for the reminder.

I think we want both. The pragma is surely a clean way, I will make an extension for this.
However, the "nondet" prefix should also have this effect, since this is exactly what the C front-end is doing. Otherwise there would be an inconsistency w.r.t. the C front-end, and users might get confused, or worse, fail to inject non-determinism without noticing.

mbeckersys:
Rebased to master, because an extension to pragmas was required (case for Annotate).

I have addressed the style comments above, and started to implement Piotr's suggested way. However, this is over my head as of now. I don't know how to safely find the pragmas belonging to the function. In the latest commit, I have drafted a quick and dirty solution to use an Annotate aspect, but I am not happy with that. 

Comments are welcome.

smowton:
I'm not sure, perhaps you could say a little about what you're trying to achieve here? I've never used the CPROVER_* magic library functions myself.

mbeckersys:
Hopefully I am not getting too much into details here. But let's see:

Actually, what I was doing here is not directly related to the CPROVER functions. Among other things, these CPROVER built-ins can be used to insert a special form of `assert`, and more importantly, there is an `assume` statement. Whenever cbmc finds one of the latter ones, it will cut out, i.e., no longer consider, executions that do not satisfy the expression. For example, some variable could be declared as follows (C code, but w:
```
void foo () {
  int x;
}
```
Without an initializer, cbmc will assume that `x` can carry any integer value. In other words, non-determinism is introduced here. However, we can use `assume` to only have cbmc consider positive values for x:
```
int x;
_CPROVER_assume (x>0);
assert(x != -2);
```
As a consequence, the assert will be verified, because it is unreachable for negative values.

Now, back to the `nondet` functions, which are another way to introduce non-determinism (most of the time you will use this, instead of creating another stack variable). Suppose you want to model a sensor read, and you cannot predict which value exactly you will get from the sensor. You could do this:
```
int8_t nondet_int8();
float adc_value = nondet_int8()*2.0;
assert (adc_value < 1000.f);
```
Now, cbmc will consider all values in [-256.f, 254.f] for `adc_value` in the assertion. What happens here is, that cbmc finds a function with prefix `nondet`, which tells it that this function can return anything in the range of its type. It will therefore not even look at the body of the function, if there is any. A forward declaration is enough.

I think we should do the same thing for Ada, to be compliant with the cbmc manual. If a user has a function with prefix `nondet`, then it is expected that this will not call a specific function that returns a single value, but it should introduce non-determinism. 

That was the entire goal of my pull request here. Instead of creating an irep for function call, which does not end up in non-determinism, we can simply place a nondet irep.

smowton:
Hmmm. I think I favour doing this on the CBMC side of the fence. I think it would be cleaner to introduce a small Ada runtime library (so e.g. we could do Cbmc.assume(...) and Cbmc.nondet()) and then replace those with appropriate CBMC constructs post-hoc. This would match how CBMC works with Java, where there is an actual Java library providing stubs (e.g. get_nondet_int() { return 0; }) which CBMC later replaces.

mbeckersys:
No problem. I also though about introducing a cbmc package (which I think you are suggesting), but then this package must be implemented somewhere for gnat to proceed. For assert and assume I think that is not necessary, since pragmas are a perfect match for that. Nondet is arguably a case for a run-time, but even more when it comes to arrays. Currently gnat2goto requires the C++ built-ins, which however do not get linked the way cbmc handles its external parser now. 

Shall I close the pull request?

smowton:
Agreed re: assert, assume, and that nondet factories ought to be in a special cbmc package which youi have to link in if you want to use it. What do you mean about C++ builtins?

mbeckersys:
I don't think nondet factories should be in a special package, since I would expect it also to work with user types, that is:
```
type bitcoin is new Float range 1.0 .. 10_000.0;
...
b := nondet_bitcoin;
```
Maybe I misunderstood, but the library will not be able to cover this, unless you want to go via generics. This is certainly more explicit and Ada-like, but also more work for the user. And also, if one expects that simply the "nondet" prefix should do the magic, then this could become a pitfall, as I described earlier. 

What I mean by C++ built-ins is this: https://github.com/smowton/cbmc/blob/external_parser_support/src/cpp/cpp_internal_additions.cpp
These kind of additions should be in the special package, for sure.

smowton:
Ah yes, see what you mean. OK for now let's do it like this; if we're having to change the prefix anyway however can we make it something less likely to accidentally clash with a real user function -- cbmc_nondet should be sufficient?

mbeckersys:
I am lost a bit, sorry. Why do we have to change the prefix anyway? That would be equally pitfall-y as ignoring "nondet". Maybe I am a bit too stubborn now, but my wish as a user is that all front-ends behave the same (implicit) when I give them equivalent programs.

The real question ought to be, how can it be done better in cbmc -- the prefix is hardcoded there. Moreover, I was really surprised when I looked at the source code and discovered that cbmc would even return a nondeterministic value, if the function has a definition (the documentation of cbmc only says: provide a declaration with "nondet" prefix, and cbmc will recognize that as request for non-determinism). So clashes are happening silently :/

If we should have a prefix for the nondet functions, then I suggest "__CPROVER_nondet", since this is what cbmc is using for most things (__CPROVER_assume, for example).

smowton:
Ahh, hadn't realised there was a documented interface to use `nondet*` for such things. Ok, in that case I'll simply merge this as-is.

PR #25, opened 2018-06-15
Handle additional constructs

danpoe:


gnat2goto/driver/tree_walk.adb
```diff
@@ -1105,6 +1127,13 @@ package body Tree_Walk is
    -------------------
 
    function Do_Expression (N : Node_Id) return Irep is
+      function Create_Dummy_Irep return Irep;
```
hannes-steffenhagen-diffblue: Maybe something like `Create_Stub_For_Missing_Implementation`? Wordy I know, but we need to take care to get these out of here again as quickly as possible.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1639,12 +1684,14 @@ package body Tree_Walk is
...
-                    Do_Expression (Low_Bound (Dsd));
-                  Bound_High : constant Irep :=
-                    Do_Expression (High_Bound (Dsd));
+                  Bound_Low  : Irep;
+                  Bound_High : Irep;
 
                begin
+                  Bound_Low := Do_Expression (Low_Bound (Dsd));
+                  Bound_High :=
+                     Do_Expression (High_Bound (Dsd));
```
hannes-steffenhagen-diffblue: This was done for debugging purposes, can you please move this back? FYI gdb understands Ada code, you can do a `catch exception` to break on an ada exception.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1779,6 +1826,11 @@ package body Tree_Walk is
          Do_Pragma_Assert_or_Assume (N_Orig, Block);
       elsif Pragma_Name (N_Orig) in Name_Annotate then
          null; -- ignore here. Rather look for those when we process a node.
+      elsif Pragma_Name (N_Orig) in Name_SPARK_Mode | Name_Global |
+      Name_Postcondition | Name_Refined_State | Name_Refined_Global |
+      Name_Precondition | Name_Loop_Invariant
+      then
+         null;
```
hannes-steffenhagen-diffblue: Leave a comment that these are ignored because they're not implemented yet, unlike the previous one which is just implemented elsewhere.

gnat2goto/driver/tree_walk.adb
```diff
@@ -2727,8 +2779,25 @@ package body Tree_Walk is
       --  Intern (Unique_Name (Corresponding_Spec (N)));
 
       Proc_Body   : constant Irep := Do_Subprogram_Or_Block (N);
-      Proc_Symbol : Symbol := Global_Symbol_Table (Proc_Name);
+      Proc_Symbol : Symbol;
    begin
+      if not Global_Symbol_Table.Contains (Proc_Name) then
```
hannes-steffenhagen-diffblue: Should probably print a warning for missing procedures. You can write to stderr with `Put_Line(Standard_Error, "text");` not sure how to get a string out of a `Proc_Name` though.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1687,10 +1687,18 @@ package body Tree_Walk is
                   Bound_Low  : Irep;
                   Bound_High : Irep;
 
+                  Dsd2 : Node_Id := Dsd;
```
hannes-steffenhagen-diffblue: Dsd2 doesn't actually seem necessary here

gnat2goto/driver/tree_walk.adb
```diff
@@ -3712,6 +3720,9 @@ package body Tree_Walk is
          when N_Pragma =>
             Do_Pragma (N, Block);
 
+         when N_Raise_Statement =>
+            null;
```
hannes-steffenhagen-diffblue: Again, we should issue a warning when ignoring things just so we don't forget we're doing it.

gnat2goto/driver/tree_walk.adb
```diff
@@ -993,6 +997,21 @@ package body Tree_Walk is
...
+   ----------------------
+
+   function Do_Real_Constant (N : Node_Id) return Irep is
+      Ret           : constant Irep := New_Irep (I_Constant_Expr);
+      Real_Constant_Type : constant Irep := Do_Type_Reference (Etype (N));
+   begin
+      Set_Source_Location (Ret, Sloc (N));
+      Set_Type (Ret, Real_Constant_Type);
+      --  ??? FIXME
+      Set_Value (Ret, "0");
```
hannes-steffenhagen-diffblue: We've already noted this down, but again for posterity: This is obviously not the correct value, we're missing an actual floating point implementation at the moment.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1126,14 +1145,15 @@ package body Tree_Walk is
    -- Do_Expression --
    -------------------
 
+   function Create_Dummy_Irep return Irep;
+   function Create_Dummy_Irep return Irep is
+      ir : constant Irep := New_Irep (I_Constant_Expr);
+   begin
+      Set_Value (ir, "0");
```
hannes-steffenhagen-diffblue: Note that these will shut up the compiler, but cbmc will choke on these things.

gnat2goto/driver/tree_walk.adb
```diff
@@ -839,6 +842,26 @@ package body Tree_Walk is
...
+   -- Do_If_Expression --
+   ------------------------
+
+   function Do_If_Expression (N : Node_Id) return Irep is
+      Expr : Node_Id := First (Expressions (N));
+      Cond : constant Irep := Do_Expression (Expr);
+      First : Irep;
+      Second : Irep;
+      Expr_Type : constant Irep := Do_Type_Reference (Etype (N));
+   begin
```
hannes-steffenhagen-diffblue: What's "First" and "Second" here? They are apparently subexpressions of the if expression, but which ones?

gnat2goto/examples/if_expr.adb
```diff
@@ -0,0 +1,5 @@
+procedure If_Expr (B : Boolean) is
```
hannes-steffenhagen-diffblue: We should add these things to the "unit" test directory (even though they're not unit tests...)

gnat2goto/driver/tree_walk.adb
```diff
@@ -830,6 +842,26 @@ package body Tree_Walk is
       return Args;
    end Do_Call_Parameters;
 
+   ------------------------
+   -- Do_If_Expression --
+   ------------------------
+
+   function Do_If_Expression (N : Node_Id) return Irep is
+      Expr : Node_Id := First (Expressions (N));
```
NlightNFotis: This `First` is a function? If yes, does this get shadowed, or is there a different namespace where this resides compared to `First` in line 852?

gnat2goto/driver/tree_walk.adb
```diff
@@ -946,9 +979,25 @@ package body Tree_Walk is
...
+            end;
+
+         when N_Package_Body =>
+            declare
+               Dummy : constant Irep := Do_Subprogram_Or_Block (U);
+               pragma Unreferenced (Dummy);
+            begin
+               Add_Start := False;
             end;
+            --  declare
```
NlightNFotis: Why are these commented out?

gnat2goto/driver/tree_walk.adb
NlightNFotis: Or maybe a significant/highlighted comment like `FIXME` or `XXX`

gnat2goto/driver/tree_walk.adb
danpoe: I think it gets shadowed, but yes I should rename it to something else.

gnat2goto/driver/tree_walk.adb
danpoe: Yeah, I'll rename them to something more descriptive.

gnat2goto/driver/tree_walk.adb
danpoe: Ah yes, this can be removed.

gnat2goto/driver/tree_walk.adb
danpoe: I think this would be hit if we encounter a subprogram for which we have not seen a declaration which would be legal ADA I think.

gnat2goto/driver/driver.adb
```diff
@@ -97,6 +100,16 @@ package body Driver is
          end loop;
       end;
 
+      if not Add_Start then
+         Put_Line (Create (SymbolTable2Json (Global_Symbol_Table)).Write);
+         return;
+      end if;
+
+      Program_Expr := New_Irep (I_Symbol_Expr);
```
smowton: If these are now only used here and below, it would be usual to declare them here using a `declare ... begin ... end` block rather than declare them above then only start using them here.

gnat2goto/driver/tree_walk.adb
```diff
@@ -830,6 +842,26 @@ package body Tree_Walk is
...
+
+   function Do_If_Expression (N : Node_Id) return Irep is
+      Expr : Node_Id := First (Expressions (N));
+      Cond : constant Irep := Do_Expression (Expr);
+      Then_Expr : Irep;
+      Else_Expr : Irep;
+      Expr_Type : constant Irep := Do_Type_Reference (Etype (N));
+   begin
+      Next (Expr);
+      Then_Expr := Do_Expression (Expr);
```
smowton: Perhaps this is the best way, but I bet there are more specific accessors for the children of `N_If_Expression`

gnat2goto/driver/tree_walk.adb
```diff
@@ -1639,12 +1727,20 @@ package body Tree_Walk is
...
-                    Do_Expression (Low_Bound (Dsd));
-                  Bound_High : constant Irep :=
-                    Do_Expression (High_Bound (Dsd));
-
+                  Bound_Low  : Irep;
+                  Bound_High : Irep;
                begin
+                  --  Produce something on which we can call Low_Bound and
+                  --  High_Bound
+                  if Nkind (Dsd) = N_Subtype_Indication then
```
smowton: Usual pattern for this: make a little function to extract the Dsd, then let the main function have the same structure as before, but call the helper (e.g. `Dsd : constant Node_Id := Get_Dsd (Spec); Bound_Low  : constant Irep := Do_Expression (Low_Bound (Dsd)); ...`)

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: Not sure how exactly this works. Ada, like C++, requires that declarations go before usage to resolve overloads and the like. Of course it counts stuffs that's in .ads files that are `with`-ed (or `new package`'ed) and I'm not sure if those end up in the global symbol table though (there should be _some_ way the frontend lets us access these though, otherwise we couldn't figure out what procedure we're actually calling in the backend).

danpoe: Yeah, currently we don't get an AST for the `.ads` file so we don't see any declarations that are in there. We should definitely improve this at some point to also process the `.ads` AST.

I think the case where we process a subprogram for which there is no symbol table entry yet might also happen when a function/procedure has been defined without a prior separate declaration.

gnat2goto/driver/tree_walk.adb
danpoe: It seems there are no specific accessors for the children. For some reason the condition, and if and else expressions are put into a list, but even if expressions with multiple `elsif` cases are parsed into nested if expressions with three operands each.

PR #27, opened 2018-06-20
Ureal to binary

hannes-steffenhagen-diffblue:
Adds a `Ureal_To_Binary_IEEE` helper that converts a "universal real" (that is GNAT lingo for rational number represented by a pair of big integers) to an IEEE_Float.

Only tested for 32 bit floats, but that's all we need for now.

Also adds a super basic unit testing framework so we can actually test things like this in isolation.

### DIFF
--- a/gnat2goto/driver/ureal_to_binary.adb
+++ b/gnat2goto/driver/ureal_to_binary.adb
@@ -1,0 +1,228 @@
+with Uintp; use Uintp;
+with Types; use Types;
+with Ada.Text_IO;
+
+package body Ureal_To_Binary is
+   package IO renames Ada.Text_IO;
+
+   procedure Get_Integer_And_Fractional_Parts (Number : Ureal;
+                                              Int_Part : out String;
+                                              Frac_Part : out String;
+                                              Int_End : out Integer;
+                                              Frac_End : out Integer)
+     with Pre => (Denominator (Number) > Uint_0);
+
+   procedure Get_Integer_And_Fractional_Parts (Number : Ureal;
+                                               Int_Part : out String;
+                                               Frac_Part : out String;
+                                               Int_End : out Integer;
+                                               Frac_End : out Integer) is
+      Num : Uint := Numerator (Number);
+      Den : constant Uint := Denominator (Number);
+
+      function Get_Next_Exponent return Int;
+      procedure Do_Integer_Part;
+      procedure Do_Fractional_Part;
+
+      function Get_Next_Exponent return Int is
+         E : Int := 0;
+         Eth_Power_Of_Two : Uint := Uint_1;
+      begin
+         while (Den * Eth_Power_Of_Two) < Num loop
+            E := E + 1;
+            Eth_Power_Of_Two := Eth_Power_Of_Two * 2;
+         end loop;
+         return E;
+      end Get_Next_Exponent;
+
+      procedure Do_Integer_Part is
+         E : Int := Get_Next_Exponent;
+         Mantissa : Uint := Uint_2**E;
+         Digit_Position : Integer := Int_Part'First;
+      begin
+         --  we start with a string filled with 0
+         Int_Part := (others => '0');
+         --  if n is the smallest integer so that
+         --  our number is < 2^n,
+         --  then we'll need n digits
+         --  (powers of 2 from 2^0 to 2^(n-1))
+         if E > Int_Part'Length then
+            raise Integer_Part_Too_Large;
+         end if;
+         --  writing to the first position
+         Int_End := Int_Part'First + Integer (E);
+         pragma Assert (Int_End <= Int_Part'Last);
+         while E >= Uint_0 loop
+            pragma Assert (Digit_Position in Int_Part'Range);
+            declare Subtrahend : constant Uint := Mantissa * Den;
+            begin
+               if Int_End > Int_Part'Last then
+                  raise Integer_Part_Too_Large;
+               end if;
+
+               if Subtrahend <= Num then
+                  Num := Num - Subtrahend;
+                  Int_Part (Digit_Position) := '1';
+               end if;
+
+               E := E - 1;
+               Digit_Position := Digit_Position + 1;
+               Mantissa := Mantissa / 2;
+            end;
+         end loop;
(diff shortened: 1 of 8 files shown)


gnat2goto/unit/test_util.adb
```diff
@@ -0,0 +1,55 @@
...
+
+      IO.Put_Line
+        ("##########################################################");
+      IO.Put_Line ("# Running suite: " & Name);
+      IO.Put_Line
+        ("##########################################################");
+      Suite.all;
+      if Suite_Failed_Test_Count > 0 then
+         IO.Put_Line ("Failed"
+                        & Positive'Image (Suite_Failed_Test_Count)
```
NlightNFotis: The attribute `Image` is essentially `to_string()` for anyone else wondering.

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -0,0 +1,228 @@
...
+                                              Frac_Part : out String;
+                                              Int_End : out Integer;
+                                              Frac_End : out Integer)
+     with Pre => (Denominator (Number) > Uint_0);
+
+   procedure Get_Integer_And_Fractional_Parts (Number : Ureal;
+                                               Int_Part : out String;
+                                               Frac_Part : out String;
+                                               Int_End : out Integer;
+                                               Frac_End : out Integer) is
```
NlightNFotis: Why are there so many out variables?

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -0,0 +1,228 @@
...
+      Num : Uint := Numerator (Number);
```
NlightNFotis: Why is `Numerator` non-constant while `Denominator` is?

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -0,0 +1,228 @@
...
+         return E;
+      end Get_Next_Exponent;
+
+      procedure Do_Integer_Part is
+         E : Int := Get_Next_Exponent;
+         Mantissa : Uint := Uint_2**E;
+         Digit_Position : Integer := Int_Part'First;
+      begin
+         --  we start with a string filled with 0
+         Int_Part := (others => '0');
```
NlightNFotis: This is array initialisation syntax, for anyone not familiar with it (I wasn't, Hannes explained)

gnat2goto/unit/floating_point_tests.adb
```diff
@@ -0,0 +1,125 @@
...
+   procedure Test_Suite;
+
+   procedure Do_Test_Suite is
+   begin
+      Test_Util.Run_Test_Suite ("Floating point literals", Test_Suite'Access);
+   end Do_Test_Suite;
+
+   procedure Test_Integer is
+      --  16/2 = 8
+      Three : constant Ureal := UR_From_Components (Uint_16, Uint_2);
```
NlightNFotis: This should be `Eight`?

gnat2goto/unit/floating_point_tests.adb
```diff
@@ -0,0 +1,125 @@
...
+        & "00000000"
+        & "0000000";
+   begin
+      pragma Assert (Convert_Ureal_To_Binary_IEEE (Three_Point_Five)
+                    = IEEE_Bits);
+   end Test_General;
+
+   procedure Test_Negative_Integer is
+      Minus_Three : constant Ureal :=
+        UR_From_Components (-Uint_12, Uint_4);
```
NlightNFotis: `Uint_12` is the constant value `12` for the `Uint` type, which is not `Unsigned Int` as it would be anticipated, but `Universal Int`

gnat2goto/unit/floating_point_tests.adb
```diff
@@ -0,0 +1,125 @@
...
+                       = IEEE_Bits);
+   end Test_Negative_General;
+
+   procedure Test_Suite is
+   begin
+
+      Uintp.Initialize;
+      Urealp.Initialize;
+
+      Test_Util.Run_Test ("Integer", Test_Integer'Access);
```
NlightNFotis: The `Access` attribute is essentially a `pointer to`.

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: Couldn't figure out a better way to return them. `String` is an indeterminate type, i.e. one that doesn't have a specific size, and thus can't be put into records (the actual definition is `type String is array(Integer range <>) of Character` IIRC), so I can't just return a record, and I wanted to return two strings (one for integer part and one for fractional) to make it easier to work with in the following functions. The `*_End` parameters could be returned, I just kept them as out parameters for symmetry (they give the index of the last meaningful character in the string).

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: The numerator isn't non-constant - Num is semantically more the "current remainder" of the operations below. It starts out with the numerator of the number, but it doesn't stay like that.

gnat2goto/unit/floating_point_tests.adb
hannes-steffenhagen-diffblue: Ohps

gnat2goto/unit/floating_point_tests.adb
```diff
@@ -0,0 +1,142 @@
...
+   procedure Test_Suite;
+
+   procedure Do_Test_Suite is
+   begin
+      Test_Util.Run_Test_Suite ("Floating point literals", Test_Suite'Access);
+   end Do_Test_Suite;
+
+   procedure Test_Integer is
+      --  16/2 = 8
+      Eight : constant Ureal := UR_From_Components (Uint_16, Uint_2);
```
danpoe: Just for clarification: So the type universal real is the type of the GNAT parsing result of some of the literals, but is not an actual ADA standard type?

hannes-steffenhagen-diffblue: Yes, this is a GNAT internal type used to represent literals and constant expressions.

PR #28, opened 2018-06-20
Floating point literals

hannes-steffenhagen-diffblue:
Adds support for floating point literals.

Depends on https://github.com/diffblue/gnat2goto/pull/25 and https://github.com/diffblue/gnat2goto/pull/27


testsuite/gnat2goto/tests/floating_point_literals/floating_point_literals.adb
```diff
@@ -0,0 +1,7 @@
+procedure Floating_Point_Literals is
+  X : Float := 1.0;
```
chrisr-diffblue: My IEEE knowledge is weak... But is there any value in also having tests with values that are fractional ?

hannes-steffenhagen-diffblue: I have written some unit tests for the ureal_to_binary conversion here: https://github.com/diffblue/gnat2goto/pull/27/files

This test is more or less just to check that this code path is actually used for float literals.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1036,8 +1037,7 @@ package body Tree_Walk is
    begin
       Set_Source_Location (Ret, Sloc (N));
       Set_Type (Ret, Real_Constant_Type);
-      --  ??? FIXME
-      Set_Value (Ret, "0");
+      Set_Value (Ret, Convert_Ureal_To_Binary_IEEE (Realval (N)));
```
chrisr-diffblue: Does this handled the difference between floating point and fixed point real types?

hannes-steffenhagen-diffblue:
Marking do not merge for a sec, forgot to include the commit that adds generation of "__CPROVER_rounding_mode"

gnat2goto/driver/driver.adb
```diff
@@ -284,4 +314,27 @@ package body Driver is
...
+   procedure Add_CProver_Internal_Symbols is
+      procedure Add_Global_Sym (Name : Symbol_Id; Sym_Type : Irep);
+      procedure Add_Global_Sym (Name : Symbol_Id; Sym_Type : Irep) is
+         Sym : constant Symbol :=
+           (Name => Name,
+            BaseName => Name,
+            PrettyName => Name,
+            SymType => Sym_Type,
+            Module => Intern (""),
+            Mode => Intern ("C"),
```
NlightNFotis: What is the symbol mode, I can't remember? Is it a hack for now to have the mode set as `"C"`, and do we plan to introduce an `"ADA"` mode, or will it be like this forever?

hannes-steffenhagen-diffblue: TBH I don't know what the mode is even good for, but as it's a cprover internal variable it should behave the same way regardless of the source language I think.

danpoe: There's also modes for C, C++, and Java (`ID_C`, `ID_cpp`, `ID_java`). AFAIK they are used to produce language-specific output when e.g. printing the symbol table or the goto program. I think we can use `"C"` for now until we run into an actual situation where we would need an ADA mode.

NlightNFotis: Very useful, thanks @danpoe

PR #35, opened 2018-08-07
Refactor ireps_generator.py

apaschos:
Remove unused variables and imports
Add missing default arguments

Part of a bigger refactoring process.

### DIFF
--- a/gnat2goto/ireps/ireps_generator.py
+++ b/gnat2goto/ireps/ireps_generator.py
@@ -28,7 +28,6 @@
 
 import os
 import os.path
-import sys
 import json
 import argparse
 
@@ -36,12 +35,29 @@ from pprint import pprint
 from glob import glob
 from copy import copy
 
+
+# Helper functions
+
 irep_type_to_ada_type = {
-   "irep"      : "Irep",
-   "bool"      : "Boolean",
-   "integer"   : "Integer",
-   "string"    : "String",
-   "gnat:sloc" : "Source_Ptr"
+    "irep"      : "Irep",
+    "bool"      : "Boolean",
+    "integer"   : "Integer",
+    "string"    : "String",
+    "gnat:sloc" : "Source_Ptr"
+}
+
+special_names = {
+    "+"      : "op_add",
+    "-"      : "op_sub",
+    "*"      : "op_mul",
+    "/"      : "op_div",
+    "**"     : "op_exp",
+    "="      : "op_eq",
+    ">"      : "op_gt",
+    "<"      : "op_lt",
+    ">="     : "op_geq",
+    "<="     : "op_leq",
+    "unary-" : "op_neg",
 }
 
 def ada_casing(s):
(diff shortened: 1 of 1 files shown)


gnat2goto/ireps/ireps_generator.py
```diff
@@ -28,22 +28,37 @@
...
-   "string"    : "String",
-   "gnat:sloc" : "Source_Ptr"
+    "irep"      : "Irep",
+    "bool"      : "Boolean",
+    "integer"   : "Integer",
+    "string"    : "String",
+    "gnat:sloc" : "Source_Ptr"
 }
 
+special_names = {"+"      : "op_add",
```
hannes-steffenhagen-diffblue: odd indentation here

apaschos: Ah well, hasty copy paste. It's already fixed in my code. I will push it with the next changes, if you don't mind.

NlightNFotis: @apaschos No, just commit the change of the fix in indentation, rebase (squash) the two commits together, then just force push and it will be fixed here.

If you fix it later, you will just be introducing irrelevant changes to another PR, and they also won't be logically coherent.

NlightNFotis: This appears to be a map of operator symbols to operator names. Can it get appropriately renamed?

gnat2goto/ireps/ireps_generator.py
```diff
@@ -623,7 +623,7 @@ def rec(sn):
 
                 friendly_name = list_schema["friendly_name"]
                 element_type  = list_schema["schema"]
-                register_sub_setter(sn, i, friendly_name, element_type, True)
+                register_sub_setter(sn, i, friendly_name, element_type, True, None)
```
NlightNFotis: Why is `None` added here as an argument, when there is no change to the function's parameters in the definition of the function?

smowton:
This was written by Florian Schanda, not me, so untagging myself

gnat2goto/ireps/ireps_generator.py
apaschos: Because it was forgotten by the original author. For some reason, python considered this fine but pylint noted this as an error.

gnat2goto/ireps/ireps_generator.py
apaschos: It's not that I cannot. It's more of not changing things that would cause the original author to find them difficult to change, since I don't know if he still bothers maintaining the file. I could have made more changes but I'm sticking to the minimum.

apaschos:
It's been rebased and the special names' indentation has been fixed.

gnat2goto/ireps/ireps_generator.py
hannes-steffenhagen-diffblue: I wouldn't worry about that; We're 'owning' this repository now. In any case, a large-ish refactoring would make any changesets based on the older version invalid most of the time anyway.

apaschos:
An issue has been raised [here](https://github.com/diffblue/gnat2goto/issues/43).

PR #36, opened 2018-08-08
Refactor continued

apaschos:
The real diff is in the last commit. What changed is I added a class and moved the big functions inside as members functions. Filtering out the whitespace should make it easy to see the changes.

### DIFF
--- a/gnat2goto/ireps/ireps_generator.py
+++ b/gnat2goto/ireps/ireps_generator.py
@@ -205,2067 +205,2071 @@ def mk_prefixed_lines(prefix, lines, join=""):
         rv.append(empty_prefix + line)
     return rv
 
-def optimize_layout(layout, max_int, max_bool):
-    accessors = {
-        "int"  : set(),
-        "bool" : set(),
-    }
-    req_fld = {
-        "int"  : max_int,
-        "bool" : max_bool,
-    }
 
-    for sn in layout:
-        for friendly_name in layout[sn]:
-            lo_typ, _, lo_kind = layout[sn][friendly_name]
-            if lo_kind in ("irep", "list"):
-                a_kind = "int"
-            elif lo_typ in ("int", "str", "sloc"):
-                a_kind = "int"
-            else:
-                lo_typ == "bool"
-                a_kind = "bool"
-            accessors[a_kind].add(friendly_name)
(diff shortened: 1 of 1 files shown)


gnat2goto/ireps/ireps_generator.py
```diff
@@ -205,2067 +205,2071 @@ def mk_prefixed_lines(prefix, lines, join=""):
...
+        schemata = {}
+        for schema_fn in schema_file_names:
+            with open(schema_fn, "rU") as fd:
+                sn = os.path.splitext(os.path.basename(schema_fn))[0]
+                # print "Loading %s" % sn
+                schemata[sn] = json.load(fd)
+                schemata[sn]["subclasses"] = set()
+                schemata[sn]["ada_name"] = ada_casing("i_" + sn)
+
+        # schemata will contain the following. Things in <> we synthesise below.
```
hannes-steffenhagen-diffblue: Why was this changed from upper to lower case?

gnat2goto/ireps/ireps_generator.py
```diff
@@ -205,2067 +205,2071 @@ def mk_prefixed_lines(prefix, lines, join=""):
...
+                            op_counts[sn]["int"] += 1
+                        else:
+                            print sn, setter_name, kind, typ
+                            assert False
+
+        MAX_INTS  = max(x["int"] for x in op_counts.itervalues())
+        MAX_BOOLS = max(x["bool"] for x in op_counts.itervalues())
+
+        if optimize:
+            layout, MAX_INTS, MAX_BOOLS = self.optimize_layout(layout,
```
hannes-steffenhagen-diffblue: This seems to have been broken in the previous version? It looks like this used to be an empty `if`, which isn't actually valid in Python.

gnat2goto/ireps/ireps_generator.py
apaschos: Accidentally. Will be fixed soon.

gnat2goto/ireps/ireps_generator.py
apaschos: I just checked it with meld locally. It seems it's a problem with github diff tool or maybe the "ignore whitespace" option. It was working before, it is also working now.

PR #41, opened 2018-08-28
Fix floating point comparison

apaschos:
Fix the erroneous generation of sub entry for floating point types.

The real diff is just the last 2 commits. One is for the change and the other one for the test of the change.

Side note: I don't know if I should do the same for when the irep is a list. I need a counterexample in order to check.


testsuite/gnat2goto/tests/floating_point_literals/test.out
```diff
@@ -1,3 +1,5 @@
 [1] : FAILURE
+[3] : FAILURE
```
hannes-steffenhagen-diffblue: This is a bit odd, normally I'd expect the successes and failures to come in order.

(Also, unrelated to this change: We will also need to get around to make the assertion format right for gnat2goto, it should read something like `[Floating_Point_Literals.assertion.1] assertion X > Z: SUCCESS`)

PR #48, opened 2018-10-19
Adding travis yaml

xbauch:


.travis.yml
```diff
@@ -0,0 +1,65 @@
...
+      make
+      gprbuild -P unit_tests.gpr -cargs -largs
+    )
+
+  - sudo apt-get install python-gnatpython
+  - export PYTHONPATH="/usr/lib/python2.7/dist-packages"
+
+script:
+  #run the tests
+  - (cd testsuite/gnat2goto; ./testsuite.py -j 2)
```
hannes-steffenhagen-diffblue: Unfortunately the testsuite won't fail with a nonzero exit code if some of the tests fail - which does happen ATM, checkout the log.

xbauch: Should we check for the result on the Travis side or change the Python script to return non-zero?

martin-cs: The test suite is deliberately similar to the GNAT one so I'd prefer it wasn't changed too much.

run_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+cd testsuite/gnat2goto
+! ./testsuite.py -j 2 2>&1 >/dev/null | tee /dev/tty | grep -q -m 1 "ERROR" && ../../gnat2goto/install/bin/unit_tests
```
hannes-steffenhagen-diffblue: I'd recommend running the integration testsuite and the unit tests independently from each other

xbauch: So we want to run all the tests and fail if there was at least one error? Or just run the two sets of tests separately and stop each one with the first error?

.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+      export GPR_PROJECT_PATH=${GNAT}/lib/gnat
+      make
+      gprbuild -P unit_tests.gpr -cargs -largs
+    )
+
+  - sudo apt-get install python-gnatpython
+  - export PYTHONPATH="/usr/lib/python2.7/dist-packages"
+
+  #install updated cbmc
+  - git clone git@github.com:NlightNFotis/cbmc.git ${CBMC}
```
NlightNFotis: I hope this is temporary; also keep in mind that I'm actively working on this, so it's shaky

.travis.yml
```diff
@@ -0,0 +1,89 @@
...
+  - export PYTHONPATH="/usr/lib/python2.7/dist-packages"
+
+  - |
+    #install updated cbmc
+    ( set -euo pipefail
+      git clone https://github.com/NlightNFotis/cbmc.git ${CBMC}
+      cd ${CBMC}
+      git checkout external_parser_support_rebased
+      ccache -z
+      ccache --max-size=1G
```
hannes-steffenhagen-diffblue: I don't think this will work, rather you'll want the `cache: ccache` option as per https://docs.travis-ci.com/user/caching/

xbauch: I'll check.

xbauch: Ok, you're right, having `cache: ccache` has the same effect as calling `ccache` explicitly.

.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+
+compiler: gcc
+
+sudo: true
+
+env:
+  GNAT: "${HOME}/opt/gnat"
+  CBMC: "${HOME}/cbmc"
+
+cache: ccache
```
chrisr-diffblue: If you look at how we handle ccache on, say, the CBMC repo you'll see theres a couple of additional things we'll want to do:
1) Specify a limit on the size of the ccache - I think for CBMC we set it to 1GB (default is something like 5GB otherwise) - this is important because over time the size of the cache will grow, and that then slows down travis downloading/uploading its cache at the beginning/end of the job
2) Its useful to zero ccaches' stats (not contents, just the counters) at the beginning of the job (`ccache -z`), and then print the ccache counters at the end of the job (`ccache -s`) - this enables us to see how effectively the ccache is working.

chrisr-diffblue: Oh, and 3) - make sure that the cbmc build is actually calling ccache, and not just calling g++... Again, check the CBMC travis job to see how that does this (IIRC it does this explicitly)

.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+      gprbuild -P unit_tests.gpr -cargs -largs
+    )
+
+  - sudo apt-get install python-gnatpython
+  - export PYTHONPATH="/usr/lib/python2.7/dist-packages"
+
+  - |
+    #build updated cbmc
+    ( set -euo pipefail
+      git clone https://github.com/NlightNFotis/cbmc.git ${CBMC}
```
chrisr-diffblue: It's probably worth adding a comment in here to make it clear that this repo/branch combination is just a temporary repo until the branch is merged. Probably also worth doing a shallow clone as well (e.g. `--depth=50`) or something like that (again, check what CBMC travis does).

chrisr-diffblue: Or possibly even better: https://stackoverflow.com/questions/1778088/how-do-i-clone-a-single-branch-in-git

.travis.yml
```diff
@@ -67,7 +67,7 @@ install:
     ( set -euo pipefail
       git clone https://github.com/NlightNFotis/cbmc.git --branch "external_parser_support_rebased" --single-branch ${CBMC}
       cd ${CBMC}
-      cmake -H. -Bbuild '-DCMAKE_BUILD_TYPE=Release'  '-DCMAKE_CXX_COMPILER=/usr/lib/ccache/g++-7'
+      cmake -H. -Bbuild '-DCMAKE_BUILD_TYPE=Release'  '-DCMAKE_CXX_COMPILER=g++-7'
```
chrisr-diffblue: Do you know what `g++-7` is on the PATH? I had a sneaky feeling that some of the travis configs put the ccache wrappers on the path before the actual compiler binaries... I could be wrong though. Specifying an absolute path is safer, and its what the CBMC travis jobs did (at least they did the last time I had to work on their travis builds).

hannes-steffenhagen-diffblue:
@martin-cs @chrisr-diffblue 
I'd consider this fairly ready at this point. If there are no further objections, I'll merge this next monday.

NlightNFotis:
@hannes-steffenhagen-diffblue Is this supposed to be failing before it gets merged? Does this get fixed by the changes in #53 ?

PR #50, opened 2018-10-25
Add a result type for explicit error handling

hannes-steffenhagen-diffblue:
This is a utility type for explicit error handling. Not used at the moment, but might be useful for some of the work we're doing right now (we might run into situations where we have to replace exceptions, pre and postconditions with explicit checks whose failures can't be immediately handled at the check site, but could be handled one or two layers above).

### DIFF
--- a/gnat2goto/driver/result.adb
+++ b/gnat2goto/driver/result.adb
@@ -1,0 +1,19 @@
+package body Result is
+
+   function Make_Error (Error_Value : Error_T) return Result_T is
+     ((Status => Error, Error => Error_Value));
+
+   function Make_Ok (Ok_Value : Ok_T) return Result_T is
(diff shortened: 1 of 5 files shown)


chrisr-diffblue:
@hannes-steffenhagen-diffblue Do you think this is still something we could/should be using? If so, could you rebase and I'll chase people for reviews. If not, could you close this PR? Just want to try to reduce the number of open PRs we have :-)

NlightNFotis:
This looks great. We should have merged it by now :)

martin-cs:
@hannes-steffenhagen-diffblue This still seems like a good idea to me, can we keep this open or are there non-technical / political / personal reasons why you would rather this is closed?

PR #51, opened 2018-10-25
Add a script to run gnat2goto against AdaYaml

hannes-steffenhagen-diffblue:
This is meant to be used to test gnat2goto feature
support against a "real" implementation. For now
we are only trying to get to the point that gnat2goto
won't crash on reaching the first unhandled node type
(and similar), so there are no tests attached


chrisr-diffblue:
I'm going to close this PR as I think it's probably been superseded by the implementation of CI_list_unsupported.sh - however if you think AdaYaml would still be a useful test program to run against, feel free to add it to that script as another program to test.

PR #65, opened 2018-11-20
Fail testsuite on travis on UOK

hannes-steffenhagen-diffblue:
An UOK in the testsuite means that the test was marked as "expected fail" XFAIL
but actually ended up passing. This is something that requires attention,
so reporting a failure in this case is appropriate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,7 @@ install:
 
 script:
   #run the tests
-  - (cd testsuite/gnat2goto; ! ./testsuite.py -j 2 2>&1 >/dev/null | tee /dev/tty | grep "ERROR" > /dev/null)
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -77,7 +77,7 @@ install:
 
 script:
   #run the tests
-  - (cd testsuite/gnat2goto; ! ./testsuite.py -j 2 2>&1 >/dev/null | tee /dev/tty | grep "ERROR" > /dev/null)
+  - (cd testsuite/gnat2goto; ! ./testsuite.py -j 2 2>&1 >/dev/null | tee /dev/tty | grep -E "ERROR|UOK" > /dev/null)
```
sonodtt: |UOK  - not |OK ?  (just checking)

PR #66, opened 2018-11-20
Feature/id nil

martin-cs:
Add the ID_nil irep type.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -2921,7 +2921,7 @@ package body Tree_Walk is
       Set_Identifier (Proc, To_String (Callee));
 
       Set_Source_Location (R, Sloc (N));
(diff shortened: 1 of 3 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -2921,7 +2921,7 @@ package body Tree_Walk is
       Set_Identifier (Proc, To_String (Callee));
 
       Set_Source_Location (R, Sloc (N));
-      --  Set_LHS (R, Empty);  -- ??? what is the "nil" irep?
+      --  Set_LHS (R, Make_Nil (Sloc (N)));
```
hannes-steffenhagen-diffblue: Should this be no longer commented now?

martin-cs: I don't know.  I didn't want to start changing things like that for now.

PR #71, opened 2018-11-22
Make bitvector output hex

hannes-steffenhagen-diffblue:
Adds a binary to hex function and makes gnat2goto use it. Function itself is easy enough and is completely reviewable, I'm not so sure if the changes in driver and tree_walk are the right changes, hence the WIP tag

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -13,7 +13,7 @@ with GNAT_Utils;            use GNAT_Utils;
 with GOTO_Utils;            use GOTO_Utils;
 with Uint_To_Binary;        use Uint_To_Binary;
 with Stand;
-with Ureal_To_Binary;       use Ureal_To_Binary;
+with Binary_To_Hex;         use Binary_To_Hex;
 with Ada.Text_IO;           use Ada.Text_IO;
 with Ada.Exceptions;
 
@@ -1187,9 +1187,10 @@ package body Tree_Walk is
       end if;
 
       Set_Source_Location (Ret, Sloc (N));
-      Set_Type (Ret, Constant_Type);
+      Set_Type (Ret, Constant_Resolved_Type);
       Set_Value (Ret,
-                 Convert_Uint_To_Binary (Intval (N), Pos (Constant_Width)));
+                 Convert_Uint_To_Hex
+                      (Intval (N), Pos (Constant_Width)));
       return Ret;
    end Do_Constant;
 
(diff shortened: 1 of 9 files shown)


hannes-steffenhagen-diffblue:
For this it's possible that all the changes to bitvector sizes are actually unnecessary and should be removed, I've not gotten around to checking that yet

hannes-steffenhagen-diffblue:
Symbol types now use the `symbol_type` ID rather than `symbol` (which is now apparently only used for variables).

hannes-steffenhagen-diffblue:
The width changes turned out to be unnecessary (thanks @xbauch for checking this out)

gnat2goto/driver/tree_walk.adb
```diff
@@ -1178,7 +1179,7 @@ package body Tree_Walk is
          null;
       else
          if Kind (Constant_Resolved_Type) in Class_Bitvector_Type then
-            Constant_Width := Get_Width (Constant_Resolved_Type);
+            Constant_Width := Get_Width (Constant_Resolved_Type) * 4;
```
chrisr-diffblue: Rather than this bit of magic arithmetic, could you add a helper function so that it has a descriptive name to indicate the purpose of the operation? e.g. `Fully_Expanded_Bitvector_Width` or something else appropriate?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1189,7 +1190,9 @@ package body Tree_Walk is
       Set_Source_Location (Ret, Sloc (N));
       Set_Type (Ret, Constant_Type);
       Set_Value (Ret,
-                 Convert_Uint_To_Binary (Intval (N), Pos (Constant_Width)));
+                 Convert_Binary_To_Hex
```
chrisr-diffblue: Is there any value in adding a `Convert_Uint_To_Hex` function to do this 'directly' - even if it's just to make the code more readable?

gnat2goto/driver/driver.adb
```diff
@@ -60,15 +60,15 @@ package body Driver is
    end GNAT_To_Goto;
 
    procedure Initialize_CProver_Internal_Variables (Start_Body : Irep) is
-      Int_32_T : constant Irep := Make_Signedbv_Type (Ireps.Empty, 8);
+      Int_32_T : constant Irep := Make_Signedbv_Type (Ireps.Empty, 32);
```
chrisr-diffblue: This commit probably needs squashing into the first commit in the series? If not, at least the parts of this commit that are undoing changes made in the first commit needs squashing.

chrisr-diffblue:
Also, the `long_integer` should be re-enabled as part of this PR.

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: This was undone in a later commit

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: Should be easy enough to do

gnat2goto/driver/driver.adb
hannes-steffenhagen-diffblue: The entire thing will be squashed once reviews are through, there's no need for this to come in multiple commits

PR #72, opened 2018-11-22
Make Travis use CBMC as a git submodule

chrisr-diffblue:


chrisr-diffblue:
OK, this is ready for review/merge. CI is failing because some of the tests seem to be hanging, which travis then times out and kills. This is unrelated to this change (it's doing that on other PRs too) so once it's had two approving reviews, I'll use admin rights to merge it.

PR #74, opened 2018-11-22
Improvements to gnat2goto CI

chrisr-diffblue:
This PR builds on #72 and adds a couple of extra improvements:

* Enable the regression tests to output a diff where tests fail due to differing output (rather than a single line saying "DIFF"
* Reduces the amount of CBMC we build, since we don't need to build all of it to be able to run the gnat2goto regression tests
* Add a 60 second timeout per-test. The default time out value is too large (760 seconds) which causes a hanging test case to then cause Travis to timeout the entire job.

Only the final two commits need to be reviewed. 
NOTE: This should not be merged until #72 has been merged

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ install:
 
 script:
   #run the tests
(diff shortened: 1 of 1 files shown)


chrisr-diffblue:
Merging even though CI is currently failing because its failing for reasons not related to this patch, and this patch will help speed up the travis jobs in general.

PR #75, opened 2018-11-22
Skip the forloop testcase because it hangs travis

chrisr-diffblue:
This test will be re-enabled as part of our work to fix up all the regression tests.


chrisr-diffblue:
Good catch Petr, thanks for spotting :-)

PR #76, opened 2018-11-22
Turn on gnat python color output for regression test results in CI

chrisr-diffblue:
NOTE, this builds on PR #75 so only the final commit in this PR is for review, and this PR will need rebasing once #75 is merged.


chrisr-diffblue:
Now rebased, but no changes so will just wait for CI to run + fail, then merge.

PR #80, opened 2018-11-26
Make testsuite.py return a fail code when any regression test fails

chrisr-diffblue:
This change to testsuite.py means we no longer need to have the fragile grep'ing of the output to indicate to travis whether any regression tests failed or not.

### DIFF
--- a/testsuite/gnat2goto/testsuite.py
+++ b/testsuite/gnat2goto/testsuite.py
@@ -14,6 +14,7 @@ from gnatpython.testdriver import add_run_test_options
 from gnatpython.reports import ReportDiff
 from glob import glob
 import os
+import sys
 
 
 def main():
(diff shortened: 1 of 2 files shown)


testsuite/gnat2goto/testsuite.py
```diff
@@ -14,11 +15,17 @@
 from gnatpython.reports import ReportDiff
 from glob import glob
 import os
+import sys
 
 
 def main():
     """Run the testsuite"""
 
+    # This is a hack that causes gantpython to treat UOK 'unexpected OK' results
```
danpoe: `gant` -> `gnat`

chrisr-diffblue:
OK, I've dropped the 'UOK' commit from this PR after all, because travis has an old version of gnatpython. Will need some more thought with that, so I'll merge this PR with just the exit code patch.

PR #81, opened 2018-11-27
Fix writing to a file instead of standard_output when translating a p…

tjj2017:
This fixes an improperly implemented change to write output to a file.
Package translations were still written to standard_out.
I identified this problem weeks ago but forgot to put in a pull request.
My later branches already have this fix in.


chrisr-diffblue:
Code changes look good to me, thanks! The PR will need rebasing on the latest master branch please, and I'll also get someone else in my team to give a second review as well.

tjj2017:
Thanks for the review.  I have rebased.

PR #82, opened 2018-11-27
Do cu specs

tjj2017:
This branch implements the processing of "withed" units.
Currently only subprogram_declarations, subprogram_bodies and package_declarations are handled.
(No generic declarations or task declarations)
There are many declarations that can be declared within these units which are not handled.

### DIFF
--- a/gnat2goto/driver/driver.adb
+++ b/gnat2goto/driver/driver.adb
@@ -87,9 +87,6 @@ package body Driver is
    procedure Translate_Compilation_Unit (GNAT_Root : Node_Id)
    is
       pragma Assert (Nkind (GNAT_Root) = N_Compilation_Unit);
-      Add_Start : Boolean;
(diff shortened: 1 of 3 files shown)


chrisr-diffblue:
@tjj2017 Many thanks for this large contribution! Could I ask for a couple of things to be done before we review it fully:

* Could you rebase it on the latest master branch
* There is currently two 'merge commits' in the list of commits for this PR - could you remove those please? (they will likely go away with the rebase, but its worth checking) We prefer not to have merge commits in PR's because it makes for a messy commit history
* Check whether any commits should be squashed together - e.g. if you have introduced diagnostic print statements in one commit, and then in a later commit removed them, those two commits should most likely be squashed together.

Thanks again for your efforts, an we'll look forward to reviewing it fully!

tjj2017:
The conflicts associated with this branch have been resolved and merged via a new branch associated with PR #113

PR #83, opened 2018-11-28
Testsuite/arrays

xbauch:
Building on #79 (thus only the last commit is new) this fixes the `arrays` test to the point that CBMC parses it. The symex takes a lot of time though.


gnat2goto/driver/tree_walk.adb
```diff
@@ -3887,6 +3901,13 @@ package body Tree_Walk is
          Func_Symbol.Mode := Intern ("C");
          Func_Symbol.Value := Body_Block;
          Global_Symbol_Table.Insert (Intern (Func_Name), Func_Symbol);
+         --  Add allocation function to symbol table
+         Alloc_Symbol.SymType := Func_Type;
+         Alloc_Symbol.Name := Intern (Alloc_Name);
+         Alloc_Symbol.PrettyName := Alloc_Symbol.Name;
+         Alloc_Symbol.BaseName := Alloc_Symbol.Name;
+         Alloc_Symbol.Mode := Intern ("C");
```
sonodtt: I don't see any use of Alloc_Symbol.Value (so no body?) - or further use of Alloc_Symbol. 
Just flagging this. 
Not enough gnat2goto experience to say if this is likely a problem going forward.

gnat2goto/driver/tree_walk.adb
```diff
@@ -3771,7 +3774,11 @@ package body Tree_Walk is
          Set_Iter (Body_Loop, Make_Increment (Counter_Sym, Index_Type, 1));
          Set_Lhs (Loop_Test, Counter_Sym);
          Set_Rhs (Loop_Test, Param_Symbol (Len_Arg));
+         Set_Type (I     => Loop_Test,
+                   Value => Make_Bool_Type);
          Set_Cond (Body_Loop, Loop_Test);
+         Set_Init (I     => Body_Loop,
+                   Value => Make_Nil (Sloc (RHS_Element_Type)));
```
sonodtt: Have not seen use of aligned name parameters elsewhere - and surprised this passes the compiler? 
If it passes, fine, but there's no precedent for this code style. 
Nitpick.

gnat2goto/driver/tree_walk.adb
```diff
@@ -3870,6 +3882,8 @@ package body Tree_Walk is
          Append_Argument (Call_Args, Param_Symbol (Ptr_Arg));
          Append_Argument (Call_Args, Param_Symbol (Len_Arg));
          Set_Arguments (Call_Inst, Call_Args);
+         Set_Lhs (I     => Call_Inst,
```
sonodtt: Lhs_fun_call is set to the LHS of the returned duplicator function. Not sure why it needs that on top of the function return value (?) It's all a bit voodoo...

gnat2goto/driver/tree_walk.adb
xbauch: Here CBMC was complaining that `__new__array` is not in the symbol table, but it only needed declaration: the definition is in the builtin functions. So I thought the body would be ignored anyway.

gnat2goto/driver/tree_walk.adb
xbauch: Not sure, but GPS provides this layout as a part of it's auto-completion. I mostly like it, but it can get very verbose sometimes.

gnat2goto/driver/tree_walk.adb
xbauch: I though about it a bit, and it sort of makes sense, i.e. to require function calls to have left-hand sides. Even if a function has a return value, if we didn't store it in the lhs, it would get lost.

sonodtt: Hmm. You're saying as a temporary. I think this _could_ therefore be implicit in the make_code_function_call bla bla function.
The return value would be assigned at the call location - or the function can be called for its side effects (temporary not needed).
But nevermind. If that's how it is - that's how it is.

xbauch: I find it unfortunate that the relevant Irep does not require the lhs given that CBMC will fail. I'm not sure if make-function-call is the only place a function can be created, but yes: you are right.

gnat2goto/driver/tree_walk.adb
chrisr-diffblue: I've seen this code style in other parts of GNAT.

testsuite/gnat2goto/tests/arrays/test.out
```diff
@@ -0,0 +1,2 @@
+VERIFICATION SUCCESSFUL
+q
```
NlightNFotis: What is `q` here?

xbauch: It's probably an emacs thing.

gnat2goto/driver/tree_walk.adb
xbauch: Empty body might do, but I'm not sure how to create one in gnat2goto. `Make_Nil` results in the same warning and `New_Irep (I_Code_Block)` breaks CBMC.

sonodtt: Ok. Thanks for extra effort + info - will look into it as part of the array epic.

sonodtt:
PR will require additional work to investigate and fix issue w.r. to error 
"raised CONSTRAINT_ERROR : Symbol_Table_Info.Symbol_Maps.Insert: attempt to insert key already in map"
Minimal reproduction: define an array of int and also an array of floats. 
Investigation WIP by @sonodtt - as part of the general 'fix array stuff' https://diffblue.atlassian.net/browse/ADA-223

gnat2goto/driver/tree_walk.adb
```diff
@@ -4186,10 +4207,39 @@ package body Tree_Walk is
...
+         Append_Declare_And_Init (Array_Copy,
+                          Make_Malloc_Function_Call (
+                            Make_Op_Mul (
+                              Lhs => Member_Size,
+                              Rhs => Param_Symbol (Len_Param),
+                              I_Type => Make_Integer_Type,
+                              Source_Location => Source_Loc)),
+                                  Body_Block, 0);
+         --  --------------------------------------------------
+         --  TYPECAST VERSION
```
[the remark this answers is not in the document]
xbauch: Well there were two versions and we wanted to get opinions as to which would be better. I removed the less generic version.

gnat2goto/driver/driver.adb
```diff
@@ -248,6 +248,7 @@ package body Driver is
 
          Global_Symbol_Table.Insert (Start_Name, Start_Symbol);
          Follow_Type_Declarations (Global_Symbol_Table, Followed_Symbol_Table);
+         --  Followed_Symbol_Table.Exclude (Intern ("malloc"));
```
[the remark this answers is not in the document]
xbauch: Removed.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1114,6 +1276,7 @@ package body Tree_Walk is
       U           : constant Node_Id := Unit (N);
       Unit_Symbol : Symbol;
    begin
+      Add_Malloc_Symbol;
```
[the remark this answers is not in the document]
xbauch: Moved.

gnat2goto/driver/tree_walk.adb
```diff
@@ -3891,70 +4174,123 @@ package body Tree_Walk is
...
+                             Source_Location => Source_Loc);
+
+         --  NEW VARS FOR MALLOC ALLOCATION
+         Member_Size : constant Irep :=
+           Make_Constant_Expr (Source_Location => Source_Loc,
+                               I_Type          => Len_Type,
+                               Range_Check     => False,
+                               Value           =>
+                    Convert_Uint_To_Hex (Value     => Esize (Element_Type) / 8,
+                                         Bit_Width => 32));
```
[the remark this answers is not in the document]
xbauch: Changed.

testsuite/gnat2goto/tests/arrays/arrays.adb
```diff
@@ -1,31 +1,7 @@
...
+   type Arr7 is array (1 .. 3) of Integer;
+   Actual7 : Arr7 := (1,2,3);
 begin
-   Actual3 (10 .. 11) := Actual3 (17 .. 18);
-   Actual3 (11) := 100;
-   -- Actual3 (13 .. 15) := 200 & (300, 400);
-   Actual3 (13 .. 16) := Actual3 (11 .. 12) & Actual3 (15 .. 16);
-   Array_Attribs := Actual3'First + Actual3'Last + Actual3'Length;
+   Actual7(1) := 2;
+--   pragma Assert(Actual7(1) =2);
```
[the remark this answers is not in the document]
xbauch: Somewhere in symex flattening. https://github.com/diffblue/gnat2goto/issues/108

xbauch:
@hannes-steffenhagen-diffblue I split it a bit more meaningfully (the last 7 commits belong here).

xbauch:
@sonodtt and @hannes-steffenhagen-diffblue this is now ready for another set of review. I've only added the malloc size fix and a failing test.

testsuite/gnat2goto/tests/arrays_access/arrays_access.adb
```diff
@@ -0,0 +1,7 @@
+procedure Arrays_Access is
+   type Arr7 is array (1 .. 3) of Integer;
```
sonodtt: rename to Arr3 and Actual3?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4299,99 +4325,120 @@ package body Tree_Walk is
...
-         Read_Ptr_Arg : constant Irep := New_Irep (I_Code_Parameter);
+      function Build_Copy_Function return Symbol;
+
+      -------------------------
+      -- Build_Copy_Function --
+      -------------------------
+
+      --  Build a functions:
+      --  n.b.: the typecast is optional
+      --  void copy_array(lhs_type* out, rhs_type* in, int len) {
```
hannes-steffenhagen-diffblue: I've been wondering - isn't this basically just `memcpy`? We do model that, for instance a C-program like this:

```
#include <string.h>
#include <assert.h>
#include <stdlib.h>

int main(void) {
  const int array_size = 10;
  int *arr = malloc(sizeof(int) * array_size);
  int *arr_copy = malloc(sizeof(int) * array_size);
  for(int i = 0; i < array_size; ++i) {
    arr[i] = i;
  }
  memcpy(arr_copy, arr, sizeof(int)*array_size);
  for(int i = 0; i < array_size; ++i) {
    assert(arr_copy[i] == i);
  }
}
```

can be verified just fine. I understand you were having troubles with this function, so maybe that can be mitigated this way? (i.e. you should be able to just call it the same way you deal with `malloc`)

testsuite/gnat2goto/tests/arrays_access/arrays_access.adb
```diff
@@ -0,0 +1,7 @@
...
+   Actual7 : Arr7 := (1,2,3);
```
hannes-steffenhagen-diffblue: Is this the bit that's triggering your problems with `array_copy` on the assign/read task? What if you leave it off, i.e. leave `Actual7` uninitialized like this: `Actual7 : Arr7;`?

testsuite/gnat2goto/tests/arrays_access/arrays_access.adb
xbauch: Done.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4409,69 +4456,100 @@ package body Tree_Walk is
...
+                                        Element_Type,
+                                        Element_Type,
+                                        Index_Type),
+                                    Lhs             => Lhs_fun_call,
+                                    Source_Location => Source_Loc);
+         Return_Inst : constant Irep :=
+           Make_Code_Return (Return_Value    => Array_Copy,
+                             Source_Location => Source_Loc);
+
+         --  NEW VARS FOR MALLOC ALLOCATION
```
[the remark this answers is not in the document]

gnat2goto/driver/tree_walk.adb
xbauch: will try, ta.

testsuite/gnat2goto/tests/arrays_access/arrays_access.adb
xbauch: It helped a bit, ta.

xbauch:
@chrisr-diffblue I've reverted the removal of the original test, adding a new, passing one.

PR #87, opened 2018-12-04
Fix typo in testsuite.py

chrisr-diffblue:


chrisr-diffblue:
It says something about Python that the instant reaction to a typo is, "Is that some sort of Python magic?" :-)

PR #98, opened 2019-01-14
Further refining of the testsuite

NlightNFotis:
This is tracking my work on fixing and reactivating some of the tests we have in light of some of the work that has been done fixing some of the issues that were causing them to be broken.

This is not there yet, I will update it accordingly when it's done.


testsuite/gnat2goto/tests/floating_point_constant/floating_point_constant.adb
```diff
@@ -0,0 +1,17 @@
+--  This test is testing the floating point constants
+procedure Floating_Point_Constant is
```
hannes-steffenhagen-diffblue: We already have `floating_point_literals`

testsuite/gnat2goto/tests/floating_point_constant/floating_point_constant.adb
```diff
@@ -0,0 +1,17 @@
...
+  --  the function here is needed because for
+  --  the second assert the constant gets simplified
+  --  into True that doesn't have a value at the moment.
+  function Haha return Integer is
+   (1);
+begin
+  if F < 5.0 then
+    pragma Assert (False);
+  end if;
+
```
sonodtt: Noticed that in previous test there is  --  CBMC: SUCCESS / FALSE prior to each check.

testsuite/gnat2goto/tests/floating_point_constant/floating_point_constant.adb
sonodtt: 1. In case relevant:
Literals and constants are significantly different things in C
https://stackoverflow.com/questions/11732795/are-the-literal-and-constant-the-same-concept-in-c

2. Should we be testing the 'constancy' aspect here?  (which a literal is, but a const integer is not necessarily a literal). 

Found this which seems interesting:
As a result, numeric literals in Ada do not carry with them their type as in C. The actual type of the literal is determined from the context.
https://www.adacore.com/gems/ada-gem-7
So the C representation of an ada literal will be ... interesting. But an Ada const?

I have only skim read enough to go: "oh!". Not enough time atm to do more.

Also:
"I'm reading Norman Cohen's Ada 95 book and on page 129 we have the constant declarations:

 Pi: constant Float := 3.1415926536
and

Pi: constant := 3.1415926536
The second declaration is said to be interpreted as, I quote: "any floating-point of fixed-point type with the appropriate range". "
https://stackoverflow.com/questions/3178339/ada-constant-declaration

So here we'd want the first. Not the second (which is a literal). 
My interpretation - from skimming.

sonodtt:
No idea why comment appeared twice!

testsuite/gnat2goto/tests/floating_point_constant/floating_point_constant.adb
hannes-steffenhagen-diffblue: Ah right, but in this case it's just us using terminology incorrectly/inconsistently. `Floating_Point_Literals` is checking that we can deal with floating point literals _and_ assign them to variables _and_ have cbmc understand expressions involving floating point stuff

testsuite/gnat2goto/tests/discriminated_record/test.out
```diff
@@ -0,0 +1 @@
+VERIFICATION SUCCESSFUL
```
hannes-steffenhagen-diffblue: Note: In this test we _mostly_ model discriminated records correctly, hence we can produce a .json_symtab without having cbmc crash on reading it. We don't really do _assignments_ on variant records correctly at the moment though.

testsuite/gnat2goto/tests/floating_point_constant/floating_point_constant.adb
sonodtt: So this 
procedure Floating_Point_Literals is
  X : Float := 1.0;

Is really just floating point number initialisation - not a 'literal' ?

"it can be said that the term literal in C stands for unnamed object that occupies memory (literals are typically lvalues)"
https://stackoverflow.com/questions/11732795/are-the-literal-and-constant-the-same-concept-in-c

PR #103, opened 2019-01-16
Fix LHS assignment in Do_Procedure_Call_Statement and Fix the recursion

NlightNFotis:
This contains the changes from #98 that are changing the behaviour of gnat2goto. This is to make #98 easier to review, as to isolate in it only the changes that were done in the test suite. After this gets merged, then #98 will be rebased on top of `master`.

### DIFF
--- a/testsuite/gnat2goto/tests/recursion/recursion.adb
+++ b/testsuite/gnat2goto/tests/recursion/recursion.adb
@@ -3,6 +3,6 @@ begin
    if (A <= 0) then
       return;
    else
-      Recursion (A - 1);
+      Recursion (A + 1);
    end if;
 end;
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -2995,7 +2995,7 @@ package body Tree_Walk is
       Set_Identifier (Proc, To_String (Callee));
 
       Set_Source_Location (R, Sloc (N));
-      --  Set_LHS (R, Make_Nil (Sloc (N)));
+      Set_Lhs (R, Make_Nil (Sloc (N)));
       Set_Function (R, Proc);
       Set_Arguments (R, Do_Call_Parameters (N));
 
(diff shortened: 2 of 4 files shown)


testsuite/gnat2goto/tests/recursion/recursion.adb
```diff
@@ -3,6 +3,6 @@ begin
    if (A <= 0) then
       return;
    else
-      Recursion (A - 1);
+      Recursion (A + 1);
```
sonodtt: ? Are we testing for non-termination?

sonodtt: ? Are we testing for non-termination?

sonodtt: ? Are we testing for non-termination?

sonodtt: I've noticed in (some) other tests there is something along the lines of 
--  CBMC SUCCESS 
prior to the check. Should we do that here?

NlightNFotis: > ? Are we testing for non-termination?

No, I had to change it to something that's already being handled in `gnat2goto`, because it doesn't handle operator minus.

> I've noticed in (some) other tests there is something along the lines of
> -- CBMC SUCCESS
> prior to the check. Should we do that here?

No, that's just an ADA comment that was documenting my expectations and I figured I should leave it there because it guides the expectations of the reader and makes it easier to verify his intuition with actual results. I don't think that's needed here if I'm honest.

hannes-steffenhagen-diffblue: @sonodtt To clarify, the test is right now just testing that we can compile a recursive procedure and have cbmc correctly deal with the generated symbol table.

In this case it turns out the problem wasn't the recursion at all, we were just doing procedure calls wrong in general.

PR #105, opened 2019-01-17
Remove problematic test results updating script

NlightNFotis:
This PR removes this script, whose job was to automatically update the expected results of tests with the current results of tests. While good as an idea initially, it ended up updating everything in an automated fashion, sometimes also changing output and marking it as expected even when it was obviously wrong.

This script ended up causing more trouble than it was adding value, therefore we decided to remove it.


martin-cs:
I'm concerned about the workflow that causes problems with this.

martin-cs:
As discussed elsewhere.

PR #111, opened 2019-01-18
Fix enum support

NlightNFotis:
Enhance enum support so that values can now be compared to each other.

### DIFF
--- a/testsuite/gnat2goto/tests/enum/enum.adb
+++ b/testsuite/gnat2goto/tests/enum/enum.adb
@@ -1,0 +1,15 @@
+procedure Enum is
+  type E is (A, B);
+
+  function Get_A return E is (A);
+  function Get_B return E is (B);
+begin
+  --  CBMC Success
+  pragma Assert (A = Get_A);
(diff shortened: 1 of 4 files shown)


testsuite/gnat2goto/tests/enum/enum.adb
```diff
@@ -0,0 +1,13 @@
...
+
+  function Get_A return E is (A);
+  function Get_B return E is (B);
+begin
+  --  CBMC Success
+  pragma Assert (A = Get_A);
+  --  CBMC Success
+  pragma Assert (B = Get_B);
+  --  CBMC Success
+  pragma Assert (Get_A /= Get_B);
```
hannes-steffenhagen-diffblue: I don't like that this test doesn't have a failing case, the way cbmc works "absurd" code like infinite loops makes assertions pass as well, so if _all_ your assertions succeed you don't really know if your code is even reachable

martin-cs: +1 yes.

PR #112, opened 2019-01-18
Initialize ada true/false variables

hannes-steffenhagen-diffblue:
The way we represent Ada Standard True / False is (similar to how we
represent enums) having them be global variables. So far unfortunately
there was no code initialising them, so things like

```
pragma Assert (True)
```

would, confusingly, lead to cbmc assertion failures. This fixes that by
initialising `standard__boolean__{true,false}` in `__CPROVER__start`

### DIFF
--- a/gnat2goto/driver/driver.adb
+++ b/gnat2goto/driver/driver.adb
@@ -121,9 +121,57 @@ package body Driver is
          end loop;
       end Initialize_Enum_Values;
 
+      procedure Initialize_Boolean_Values;
+      procedure Initialize_Boolean_Values is
+         True_Symbol_Expr : constant Irep := Make_Symbol_Expr
+           (I_Type => Make_Bool_Type,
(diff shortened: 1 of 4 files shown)


martin-cs:
@hannes-steffenhagen-diffblue : why are they handled like enums / as global variables and not translated to the appropriate constants?

hannes-steffenhagen-diffblue:
@martin-cs That would've been an option as well, but that would've required either special case handling every time we read a symbol in `Tree_Walk` or otherwise do a walk over all functions in the symbol table after tree walk. I thought this would be less confusing.

So, not a technical limitation, it was just obvious to me if one was 'better' somehow than the other and this is easier/cleaner to implement.

martin-cs:
@hannes-steffenhagen-diffblue OK; I will trust your judgement on this.  If it turns out that it is an issue then it doesn't look like it will be an excessive amount of work to change it.

PR #113, opened 2019-01-22
Do cu specs

xbauch:
Mostly Trevor's work, the last commit addresses some issues from https://github.com/xbauch/gnat2goto/pull/2 and adds a test for packages.


gnat2goto/driver/tree_walk.adb
```diff
@@ -3729,6 +3780,54 @@ package body Tree_Walk is
...
+   -------------------------
+
+   procedure Do_Withed_Unit_Spec (N : Node_Id) is
+      Unit_Name : constant String := Get_Name_String (Get_Unit_Name (N));
+   begin
+      if Defining_Entity (N) = Stand.Standard_Standard or else
+        Unit_Name = "system%s"
+      then
+         null;
+         --   At the moment Standard or System are not processed - to be done
```
sonodtt: Maybe some Put_Line warning about this?
In the message not sure if 'package' or 'unit' might be more appropriate
Put_Line (Standard_Error,
                         "At the moment Standard or System library units are not processed. ");

gnat2goto/driver/tree_walk.adb
```diff
@@ -3729,6 +3780,54 @@ package body Tree_Walk is
...
+      else
+
+         case Nkind (N) is
+            when N_Subprogram_Body =>
+               if Acts_As_Spec (N) then
+                  --  The unit is a withed library unit which subprogram body
```
sonodtt: suggest add summary line at top (similar to this)
--  withed library unit that has no separate declaration
(this gives the reader what is common among the scenario discussed below)

gnat2goto/driver/tree_walk.adb

gnat2goto/driver/tree_walk.adb

xbauch:
@sonodtt and @hannes-steffenhagen-diffblue I've addressed the comments made here and on https://github.com/xbauch/gnat2goto/pull/2. I have added commit messages summarising their content.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4312,10 +4307,11 @@ package body Tree_Walk is
...
         Unit_Name = "system%s"
       then
-         null;
-         --   At the moment Standard or System are not processed - to be done
+         --  At the moment Standard or System are not processed: TODO
+         Put_Line (Standard_Error,
+         "At the moment Standard or System library units are not processed.");
       else
-
+         --  Withed library unit that has no separate declaration
```
tjj2017: I do not understand this comment. These withed library units do have library-level declarations.  Any user defined library-level unit must have an explicit declaration whether it is a package, subprogram declaration or just a subprogram body.  This is what I tried to explain in the comments under each case.

A more general comment preceding the case statement might be
"--  Handle all other withed library unit declarations"

gnat2goto/driver/tree_walk.adb
```diff
@@ -204,6 +207,12 @@ package body Tree_Walk is
    function Do_Op_Not (N : Node_Id) return Irep
    with Pre => Nkind (N) in N_Op;
 
+   procedure Do_Package_Declaration (N : Node_Id)
```
tjj2017: I seemed to have introduced a number of procedures starting with Do_* which is not in accordance with the naming convention according to https://github.com/xbauch/gnat2goto/pull/2#discussion_r249712062.

I don't think its worth changing now because it should mean re-ordering the declarations again to keep them in alphabetical order.  If it is ok with everyone merge with these names now and I will sort out the naming conventions when I do the next PR.

gnat2goto/driver/tree_walk.adb
xbauch: I take it back: not processing standard/system libraries is not a problem unless you rely on something in them (and then you would get the appropriate error message).

gnat2goto/driver/tree_walk.adb
xbauch: Agreed, changed.

gnat2goto/driver/tree_walk.adb
sonodtt: @tjj2017 I have a small number of queries that are also just minor nitpicks / request for clarification (often just due to my poor knowledge of ADA) that I think should not get in the way of merging what is a great body of work. 
I'll ask about these separately - if additional comments are helpful then they can be in a separate minor PR.

gnat2goto/driver/tree_walk.adb
sonodtt: @xbauch Nitpick:
Okay. The "if then else" construct begs the question. 
To prevent another reader wondering why the null branch of the "if then else" rather than just just "if not" maybe add a clarifying line?
I'll assume the if else then construct is in case in the future we want to consider doing something with standard/system libraries.

gnat2goto/driver/driver.adb
```diff
@@ -226,6 +226,17 @@ package body Driver is
       end;
 
       if not Add_Start then
```
sonodtt: Default mode "don't add comments" to this thing I had to figure out what it is doing by looking at the the out parameter in another function. 
Status: "on".  
:)

PR #124, opened 2019-02-12
Generalised array assignment

xbauch:
The first commit is not part of this PR. The rest rebuilds array assignment assuming that the rhs is a concatenation.


gnat2goto/driver/tree_walk.adb
```diff
@@ -1007,109 +1008,254 @@ package body Tree_Walk is
    -- Do_Array_Assignment --
    -------------------------
 
-   --  TODO: multi-dimensional arrays
-   function Do_Array_Assignment (LHS_Expr : Irep; RHS_Expr : Irep; N : Node_Id)
-                                return Irep
+   --  The following function builds a generalised array assignment
+   --  dest := src_1 & src_2 & .. & src_n
+   --  for $n>0$, src_i may overlap with dest
```
sonodtt: I interpret this statement as being trivially true (n must be greater than 0 or you have no source and no assignment)
even in case of the identity assignment, your source would exist. 
Perhaps in this case less is more:
"src_i may overlap with dest" or "src_i may be part of dest"

gnat2goto/driver/tree_walk.adb
```diff
@@ -1007,109 +1008,254 @@ package body Tree_Walk is
...
+   --  Let ArrT := struct { int first; int last; T* data; }
+   ----------------------------------------------------------------------------
+   --  void concat_assign(ArrT dest, ArrT src_1, ArrT src_2, .., ArrT src_n) {
+   --    offset_step = 0;
+   --    slice_size = src_1.last - src_1.first + 1;
+   --    slice_temp = (T*)malloc(slice_size * sizeof(T));
+   --    memcpy(slice_temp, src_1.data, slice_size * sizeof(T));
+   --    memcpy(dest.data + offset_step, slice_temp, slice_size * sizeof(T));
```
sonodtt: I'd suggest clarifying that the temporary is required due the scenario of an array copying a slice back to itself

testsuite/gnat2goto/tests/arrays_concat/arrays_concat.adb
```diff
@@ -0,0 +1,8 @@
+procedure Arrays_Concat is
+   type Arr is array (1..4) of Integer;
+   Actual : Arr := (1,2,3,4);
+begin
+
+   Actual (1 .. 4) := Actual (3 .. 4) & Actual (1 .. 2);
+   pragma Assert (Actual(2)=4);
+end Arrays_Concat;
```
sonodtt: As I understand it the new array assignment function generalises not just to self assignment but also
A = B + C 
if (A has size 4) and B and C have size two. 

You do the edge case of self assignment - that's essential (i.e. the memcpy temp array required)
But I'd find it helpful to see *in the test* that the general case is also covered. 

To someone, in years gone by, looking at these tests in isolation the thought might occur:
"Ok - I see this covers the edge case - do I know for sure that it works with the general case?".

gnat2goto/driver/tree_walk.ads
```diff
@@ -76,4 +76,5 @@ package Tree_Walk is
      return Symbol
    with Pre => Nkind (N) = N_Compilation_Unit;
 
+   type Irep_Array is array (Positive range <>) of Irep;
```
sonodtt: This is an example of a scenario where a commit comment as to why this type was introduced would have been really helpful. 
A reviewer must now look at the locations where this is used and reverse engineer why it is needed / intent.

xbauch: Added.

gnat2goto/driver/tree_walk.adb
```diff
@@ -771,27 +771,20 @@ package body Tree_Walk is
...
-         Append_Op (Result_Block,
-                    Make_Code_Assign (Rhs             => Malloc_Call_Expr,
-                                      Lhs             => Data_Mem_Expr,
-                                      Source_Location => Source_Loc));
+         --  As long as symex is field-insensitive we need to initialise the
+         --  array structure with the information about allocated size.
+         Append_Declare_And_Init (Symbol     => Array_Temp,
+                                  Value      => Array_Temp_Struct,
+                                  Block      => Result_Block,
+                                  Source_Loc => Source_Loc);
```
sonodtt: "create a temporary struct and assign it in one swoop to Array_Temp - so that Symex does not see the struct as having been changed after its creation and can therefore see it as constant - which means that the allocated size become visible"

Is this what you meant?
If so - I think you left this small bit out...  (which would however be rather vital for the comprehension of what you're doing here).  

Pass this information on to the next person please... or you're making life rather difficult for who follows you.

gnat2goto/driver/tree_walk.adb
xbauch: Yes, the $n>0$ relates to the line above, which I guess is confusing. I'll rewrite it.

gnat2goto/driver/tree_walk.adb
xbauch: Added.

testsuite/gnat2goto/tests/arrays_concat/arrays_concat.adb
xbauch: Hmm, I thought about the (potentially) overlapping case as the general one, but I can add a test for `A=B&C`.

gnat2goto/driver/tree_walk.adb
xbauch: I've added your message to the commit as well as the comment above this initialisation.

xbauch:
@sonodtt I've addressed your comments and added a test-case.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1007,113 +984,311 @@ package body Tree_Walk is
...
 
-      return Ret;
+   -------------------------
+   -- Do_RHS_Array_Assign --
+   -------------------------
 
-   end Do_Array_Assignment;
+   function Do_RHS_Array_Assign (N : Node_Id) return Irep_Array
+   is
+   begin
```
hannes-steffenhagen-diffblue: I'm not sure I like the design of this. Basically, this creates a special case just for op-concat, and we can't use this in `Do_Expression` because it invokes `Do_Expression` itself?

gnat2goto/driver/tree_walk.adb
```diff
@@ -3634,7 +3539,8 @@ package body Tree_Walk is
    function Do_Operator_General (N : Node_Id) return Irep is
    begin
       if Nkind (N) = N_Op_Concat then
-         return Do_Op_Concat (N);
+         return Report_Unhandled_Node_Irep (N, "Do_Operator_General",
+                                            "Concat unsupported");
```
hannes-steffenhagen-diffblue: ??? Wasn't the point of this to support `Op_Concat`?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4144,53 +4050,109 @@ package body Tree_Walk is
...
    --------------
 
+   --  The following build an expression representing slice
+   --  orig(start .. end)
+   --  Let ArrT := struct { int first; int last; T* data; }
+   ----------------------------------------------------------------------------
+   --  ArrT slice_expr(ArrT orig) {
+   --    T* new_data = data + (start - orig.first);
+   --    ArrT temp_array = {.first=start, .last=end, .data=new_data};
+   --    return temp_array;
```
hannes-steffenhagen-diffblue: Repeating the question about null slices here (not sure where the right place to handle those is actually, but this looks like one place to look at?)

testsuite/gnat2goto/tests/arrays_assign_nonoverlap/arrays_assign_nonoverlap.adb
```diff
@@ -0,0 +1,9 @@
+procedure Arrays_Assign_Nonoverlap is
+   type Arr_Long is array (1..4) of Integer;
+   Full_Arr : Arr_Long := (1,2,3,4);
+   Part1_Arr : Arr_Long := (1,2);
```
hannes-steffenhagen-diffblue: ??? This isn't actually valid Ada though? `Arr_Long` is a size 4 array, you can't initialise it with a size 2 aggregate (it's syntactically valid but would raise an exception at runtime if we supported exceptions).

gnat2goto/driver/tree_walk.adb
```diff
@@ -1007,113 +984,311 @@ package body Tree_Walk is
...
-   function Do_Array_Assignment (LHS_Expr : Irep; RHS_Expr : Irep; N : Node_Id)
-                                return Irep
+   --  The following function builds a generalised array assignment
+   --  dest := src_1 & src_2 & .. & src_n         for $n$ greater or equal to 1
+   --  where each src_i may overlap with dest
+   --  (which is why we copy each src_i to a temporary before copying to dest)
+   --  Let ArrT := struct { int first; int last; T* data; }
+   ----------------------------------------------------------------------------
+   --  void concat_assign(ArrT dest, ArrT src_1, ArrT src_2, .., ArrT src_n) {
+   --    dest_temp = (T*)malloc(sum_size * sizeof(T));
```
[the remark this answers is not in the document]
xbauch: Yes it would. Fixed.

gnat2goto/driver/tree_walk.adb
xbauch: Well we could (and will in the next PR) use it in the `Do_Expression` but now it returns an array of ireps and `Do_Expression` expects just one.

gnat2goto/driver/tree_walk.adb
xbauch: Not as a general operator but only as the rhs of array assignment. The extension is relatively simple and will be in the next PR.

testsuite/gnat2goto/tests/arrays_assign_nonoverlap/arrays_assign_nonoverlap.adb
xbauch: Ok, fixed.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1007,113 +984,311 @@ package body Tree_Walk is
...
+   --    offset_step = 0;
+   --    slice_size = src_1.last - src_1.first + 1;
```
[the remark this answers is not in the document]
xbauch: Works (probably). I made it part of the `arrays_assign_nonoverlap` test.

gnat2goto/driver/tree_walk.adb
xbauch: As above.

xbauch:
@hannes-steffenhagen-diffblue Maybe resolved your comments. I guess the biggest question is if we mind leaving the concat in general_op for later.

gnat2goto/driver/tree_walk.ads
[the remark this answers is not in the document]
NlightNFotis: Ahh, I see now in the next commit why we need that. Perhaps this commit should be squashed with the next one? Otherwise this doesn't make a lot of sense.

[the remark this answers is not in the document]
xbauch: Squashed.

testsuite/gnat2goto/tests/arrays_write_element/test.out
```diff
@@ -1,4 +1 @@
-[precondition_instance.1] memcpy src/dst overlap: SUCCESS
-[precondition_instance.2] memcpy source region readable: SUCCESS
-[precondition_instance.3] memcpy destination region writeable: SUCCESS
 VERIFICATION SUCCESSFUL
```
chrisr-diffblue: Why have these expected assertion successes been removed? What does the arrays_write_element test actually assert, and why are the assertion successes/etc not expected in the output? Or is this whole test now superseeded by other tests?

xbauch: @chrisr-diffblue Assertions were removed because the goto function no longer contains the memcpy call, so no checks are generated for it. The test does not assert anything, just tries to write into an array using an index (CBMC never gets in symex). And yes: it is contained in other tests.

PR #125, opened 2019-02-13
Make unsupported feature listing script more robust

xbauch:
In accordance to ADA-249 (and ASVAT issue #33) we add inclusion of some Ada
paths as well as checking that gnat2goto can list features. We also fail
early when gnat2goto fails (in an unexpected way).

### DIFF
--- a/experiments/list_unsupported.sh
+++ b/experiments/list_unsupported.sh
@@ -34,8 +34,32 @@ done
 
 echo "$1: Unsupported features\n" > "$file_name".txt
 
+if ! command -v gnat > /dev/null; then
+   echo >&2 "Gnat not on PATH!"
+   exit 4
+fi
+
+ADA_HOME=`which gnat`
+ADA_HOME="$(dirname "$ADA_HOME")/.."
+PLATFORM=`uname -m`
+DEF_ADA_INCLUDE_PATH="${ADA_HOME}/lib/gcc/${PLATFORM}-pc-linux-gnu"
+ADA_VERSION=$(find ${DEF_ADA_INCLUDE_PATH} -maxdepth 1 -type d -name '*.*.*' -print -quit)
+DEF_ADA_INCLUDE_PATH="${ADA_VERSION}/rts-native/adainclude"
+export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-$DEF_ADA_INCLUDE_PATH}"
+
+export GPR_PROJECT_PATH="${GPR_PROJECT_PATH:-/opt/gnat/lib/gnat}"
+if [ ! -d "$GPR_PROJECT_PATH" ]; then
+   echo >&2 "GPR project path does not exists!"
+   exit 6
+fi
+
 for filename in $(find ${path} -name *.adb); do
    gnat2goto ${include_path} "${filename}" >>"$file_name".txt 2>&1
+   if [ "$?" -gt 0 ]; then
+      echo >&2 "Errors occurred during feature collection."
+      echo >&2 "See \"${file_name}.txt\" for details."
+      exit 7
+   fi
 done
 
 sed '/^\[/ d' < "$file_name".txt > "$file_name"_redacted.txt


experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
 
 echo "$1: Unsupported features\n" > "$file_name".txt
 
+export ADA_INCLUDE_PATH=/opt/gnat/lib/gcc/x86_64-pc-linux-gnu/4.9.4/rts-native/adainclude/
```
chrisr-diffblue: This would prevent the user of the script from setting their own path if they have tools installed in a different location. Probably better to do something like:
```
export ADA_INCLUDE_PATH=${ADA_INCLUDE_PATH:-/opt/gnat/lib/gcc/x86_64-pc-linux-gnu/4.9.4/rts-native/adainclude/}
```
(I've not checked that syntax, so you may need to check quoting of the path...)

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
...
+if [ ! -d "$ADA_INCLUDE_PATH" ]; then
+   echo >&2 "Ada include path does not exists!"
+   exit 4
+fi
+
+PATH=/opt/gnat/bin:${PATH}
```
chrisr-diffblue: I probably wouldn't explicitly set the path here, but instead just do the test for gnat being on the path and leave it up to the user to change their path if needs be.

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
...
+if ! command -v gnat > /dev/null; then
+   echo >&2 "Gnat not on PATH!"
+   exit 5
+fi
+
+GPR_PROJECT_PATH=/opt/gnat/lib/gnat
```
chrisr-diffblue: I'd do the same as ADA_INCLUDE_PATH here - i.e. only set this value if the user hasn't already set it.

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
...
+if [ ! -d "$GPR_PROJECT_PATH" ]; then
+   echo >&2 "GPR project path does not exists!"
+   exit 6
+fi
+
 for filename in $(find ${path} -name *.adb); do
    gnat2goto ${include_path} "${filename}" >>"$file_name".txt 2>&1
+   if [ `echo $?` -gt 0 ]; then
```
chrisr-diffblue: Don't need to `echo $?`, just use `$?` on it's own - its a variable :-)

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
...
+      echo >&2 "Errors occurred during feature collection."
```
chrisr-diffblue: Assuming gant2goto failed with a non-zero exit code, does the log of what failed get written anywhere? If so, it's worth the error message saying something like "see /foo/blah/wibble/log.txt for details"

experiments/list_unsupported.sh
xbauch: But what if neither the user-defined path nor our path exists? I rewrote it a bit differently, but can do your way as well. Your choice.

experiments/list_unsupported.sh
xbauch: Done.

experiments/list_unsupported.sh
xbauch: As above.

experiments/list_unsupported.sh
xbauch: And previously I even stored the echo in a new variable, just to be sure.

experiments/list_unsupported.sh
xbauch: Done.

xbauch:
@chrisr-diffblue I addressed your comments (maybe not as intended).

experiments/list_unsupported.sh
chrisr-diffblue: They idiomatic way to do this in most sh scripting is:
```
export FOO={FOO:-"default value for FOO if not already set"}
if [ FOO is not valid ] ; then
   error stuff
fi
```
So I'd suggest doing it that way.

xbauch: Done.

xbauch:
@chrisr-diffblue reworked and rebased.

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
...
 for filename in $(find ${path} -name *.adb); do
    gnat2goto ${include_path} "${filename}" >>"$file_name".txt 2>&1
+   if [ "$?" -gt 0 ]; then
+      echo >&2 "Errors occurred during feature collection."
+      echo >&2 "See \"${file_name}.txt\" for details."
+      exit 7
+   fi
 done
 
 sed '/^\[/ d' < "$file_name".txt > "$file_name"_redacted.txt
```
chrisr-diffblue: This is not related to this PR... But have am I right that currently this script will only report the unsupported features for the *last* file it processes, and not *all* the files it processes? In particular, we loop through all the files running gant2goto, outputting to a `${file_name}.txt` file - so one output file per input file... But we only redact and CollectUnsupported *outside* the loop....

xbauch: I should double check but I think that we append to `file_name` the result of running gnat2goto on each `filename`. Perhaps not the best choice of variable names.

chrisr-diffblue: Ah, I see - that's a relief :-) I confused 'filename' and 'file_name'.

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,30 @@ done
 
 echo "$1: Unsupported features\n" > "$file_name".txt
 
+export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-/opt/gnat/lib/gcc/x86_64-pc-linux-gnu/4.9.4/rts-native/adainclude}"
```
hannes-steffenhagen-diffblue: This still has a hard coded path; I don’t particularly like these, it makes it too easy for scripts to work by accident. Default paths should only be used if there is a reasonable expectation that something can actually be found at that path

xbauch: I've done something to improve this.

xbauch:
@hannes-steffenhagen-diffblue What do you think about this version?

chrisr-diffblue:
Can we please have @martin-cs review this too, before we merge.

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,35 @@ done
...
 echo "$1: Unsupported features\n" > "$file_name".txt
 
+if ! command -v gnat > /dev/null; then
+   echo >&2 "Gnat not on PATH!"
+   exit 4
+fi
+
+ADA_HOME=`which gnat`
+ADA_HOME="$(dirname "$ADA_HOME")/.."
+DEF_ADA_INCLUDE_PATH="${ADA_HOME}/lib/gcc/x86_64-pc-linux-gnu"
```
martin-cs: No chance of a `uname -m` instead of the hard coded `x86_64` is there?

experiments/list_unsupported.sh
```diff
@@ -34,8 +34,35 @@ done
...
+ADA_VERSION=$(find ${DEF_ADA_INCLUDE_PATH} -maxdepth 1 -type d -name '*.*.*' -print -quit)
+DEF_ADA_INCLUDE_PATH="${ADA_VERSION}/rts-native/adainclude"
+export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-$DEF_ADA_INCLUDE_PATH}"
+if [ ! -d "$ADA_INCLUDE_PATH" ]; then
```
chrisr-diffblue: @martin-cs - There's been a bit of debate today about this particular check here, because it's perfectly valid for ADA_INCLUDE_PATH to be a list of directories, not a single directory. Two options are (a) drop this test, and just check that the variable is set, (b) somehow parse the include path variable, then check each element in the list - and also handle the case that it's also valid to have some elements in the list be non existent.... We were starting to lean towards (a).

experiments/list_unsupported.sh
chrisr-diffblue: Good point!

xbauch: Agreed. Modified.

martin-cs: Ta.

experiments/list_unsupported.sh
xbauch: I removed the check for now.

xbauch:
@chrisr-diffblue @martin-cs Fixed based on comments and squashed.

experiments/list_unsupported.sh
martin-cs: (a). check that it is set but don't try to parse.

The variable is meaningful to GNAT and to the user; we don't need to parse it.  Too much cleverness here could bite us later.  Also it is valid for the path to contain empty or non-existent directories (as long as it does have some of ones that work).

PR #132, opened 2019-02-25
Add failing test for unsupported declarations

xbauch:
This new tests only covers N_Renaming_Declaration, N_Block_Statement, and
N_Use_Type_Clause. The first two fail in gnat2goto with warning and the third
fails in CBMC with malformed symbol.


hannes-steffenhagen-diffblue:
Do you need all three of these at once in one test to make this work?

PR #133, opened 2019-02-25
Add failing test for funny real literals

xbauch:
We do not support negative denominators which can only happen for real literals
represented with base 0. Small reals are represented with base 2 for which
denominator cannot be negative. I think the check can be safely removed but
perhaps we want to investigate the consequences.

### DIFF
--- a/testsuite/gnat2goto/tests/real_literal/real_literal.adb
+++ b/testsuite/gnat2goto/tests/real_literal/real_literal.adb
@@ -1,0 +1,5 @@
+procedure Real_Literal is
+  X : Float := 16777215.0*(2.0**104);
+begin
+   pragma Assert (X>0.0);
+end Real_Literal;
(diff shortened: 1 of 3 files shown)


hannes-steffenhagen-diffblue:
What's a base 0 denominator?

NlightNFotis:
I want to approve this but I don't understand where the check is located in the code so we get some context?

hannes-steffenhagen-diffblue:
So about the base 0 thing: So this has to do with the `ureal` to binary conversion. I had some checks in here because I was assuming the denominator should always be > 0, turns out that `ureal` has a second representation where this isn't always the case. In the implementation I use the _normalised_ version of the `ureal`, for which this _is_ always true, so I think these checks can genuinely just be removed.

xbauch:
The check is a prerequisite to `Get_Integer_And_Fractional_Parts`  at `ureal_to_binary.adb:11` and propagated to `tree_walk.adb:1612`.

NlightNFotis:
Now that I read Hannes' comment it makes sense.

PR #134, opened 2019-02-26
Add failing test for character literals

xbauch:
We don't support character literals: this needs a new function (similar to
`do_constant`) to be called from `do_expression`.

### DIFF
--- a/testsuite/gnat2goto/tests/character_literal/character_literal.adb
+++ b/testsuite/gnat2goto/tests/character_literal/character_literal.adb
@@ -1,0 +1,6 @@
+procedure Character_Literal is
+   A : Character := 'a';
+   B : Character := 'b';
+begin
+   pragma Assert (A > B);
+end Character_Literal;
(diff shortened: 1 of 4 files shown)


testsuite/gnat2goto/tests/character_literal/character_literal.adb
```diff
@@ -0,0 +1,6 @@
+procedure Character_Literal is
+   A : Character := 'a';
+   B : Character := 'b';
+begin
+   pragma Assert (A > B);
```
sonodtt: If not essential, I'd suggest removing the pragma. It's not a requirement of the failure case.

xbauch: Will remove, it doesn't make sense here.

hannes-steffenhagen-diffblue: FWIW I think it's worth keeping it in here as it'll require less changes when converting this to a real case

sonodtt: @hannes-steffenhagen-diffblue  
If so: "If you keep it - I'd suggest a pragma that would pass".   No?

Issue #141, opened 2019-03-06
Sort files to run as tests alphabetically


NlightNFotis:
In the test library written in python, we call `os.listdir()` to enlist the `*.adb` files in a folder to run for tests. Its documentation suggests that it returns items in an arbitrary order. 

https://github.com/diffblue/gnat2goto/blob/168240f0eb26c713a6242ae2a06f72395663d8d6/testsuite/gnat2goto/lib/python/test_support.py#L20

Perhaps we should sort it so that tests that happen to run multiple files in the same folder, like `op_and/` or `op_or/` don't accidentally break on a different machine because the function might return a different order that doesn't match the execution order expected in `test.out` in that folder.

PR #142, opened 2019-03-06
Fix handling of operator or.

NlightNFotis:
Before this change, the operator or had been implemented as the short circuiting operator, which in reality is or else. This fixes the semantics to be closer to the non-short circuiting operator.


NlightNFotis:
This is failing CI because of the issue described in #141 . Hannes will write a PR to fix it, and after it gets merged, I will rebase and fix any issues caused.

PR #147, opened 2019-03-06
Simplify the two functions that handle or and and operators

NlightNFotis:
Instead of having two very similar functions that do the same thing but differ in one line, provide one polymorphic function that calls the appropriate function through a function pointer passed to it. I think this should simplify maintenance in the future.

Also add some documentation.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -230,10 +230,18 @@ package body Tree_Walk is
    function Do_Op_Minus (N : Node_Id) return Irep
    with Pre => Nkind (N) in N_Op;
 
-   function Do_Op_Or (N : Node_Id) return Irep
-   with Pre => Nkind (N) in N_Op;
+   type Bit_Operand_Constructor is
+     access function (Lhs : Irep;
+                      Rhs : Irep;
+                      Source_Location : Source_Ptr;
+                      Overflow_Check : Boolean;
+                      I_Type : Irep;
+                      Range_Check : Boolean)
+                     return Irep;
 
-   function Do_Op_And (N : Node_Id) return Irep
+   function Do_Bit_Op (N : Node_Id;
+                       Operator : Bit_Operand_Constructor)
+                      return Irep
    with Pre => Nkind (N) in N_Op;
 
    procedure Do_Package_Declaration (N : Node_Id)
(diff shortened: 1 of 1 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -3531,36 +3539,26 @@ package body Tree_Walk is
...
 
-   function Do_Op_And (N : Node_Id) return Irep is
+   --  We identified that the constructor for operator `or` and operator
+   --  `and` where pretty much the same, with the only difference being
+   --  the constructor being called. So to avoid needless duplication,
+   --  we simplified it to a single function that does the same thing,
+   --  and just calls the appropriate constructor via a function pointer.
+   --
+   --  This produces the following code in pseudocode
+   --    A or B  (assume A => True and B => True)
```
hannes-steffenhagen-diffblue: I feel this is a bit confusing - this looks like it'd be the pseudocode, but it's actually still part of the 'preamble' so to speak. How about 

> let A and B be `True` in the following example (pseudocode)

gnat2goto/driver/tree_walk.adb
```diff
@@ -3531,36 +3539,26 @@ package body Tree_Walk is
...
+   --  --
+   --    int intA = (int) A;       // 1
+   --    int intB = (int) B;       // 1
+   --    int intC = A | B          // bitwise `or` or `and`
```
hannes-steffenhagen-diffblue: I'd just say `bitop` instead of `|` followed by `// bitop can be bitor, bitand`

and theoretically also `xor` and things like that.

gnat2goto/driver/tree_walk.adb
```diff
@@ -3531,36 +3539,26 @@ package body Tree_Walk is
...
-   end Do_Op_Or;
-
-   -------------------------
-   --      Do_Op_And      --
+   --      Do_Bit_Op      --
    -------------------------
 
-   function Do_Op_And (N : Node_Id) return Irep is
+   --  We identified that the constructor for operator `or` and operator
+   --  `and` where pretty much the same, with the only difference being
```
xbauch: where -> were

gnat2goto/driver/tree_walk.adb
NlightNFotis: Good point.

hannes-steffenhagen-diffblue: Also, the actual generated code looks more like `(bool) ((int) A bitop (int) B)`?

NlightNFotis: Yes, indeed, I wanted to make it as simple as possible, and try to simulate the sequence of the calls happening inside the function.

PR #148, opened 2019-03-07
Introduce floating point type definition

xbauch:
The type definition may contain ranges so this PR also brings in a new irep spec for bounded floatbv type, conversion to regular floatbv type, storage for real bounds, etc. We also found a bug in the conversion and demonstrate it via a modified test.

### DIFF
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -329,6 +329,32 @@ package body GOTO_Utils is
                           I_Type          => Pointer_Type);
    end Offset_Array_Data;
 
+   function To_Float_Format (Float_Type : Irep) return Float_Format
+   is
+      Float_Width : constant Integer := Get_Width (Float_Type);
+      Unsupported_Float_Width : exception;
+   begin
+      case Float_Width is
+         when 32 => return IEEE_32_Bit;
+         when 64 => return IEEE_64_Bit;
+         when others =>
+            raise Unsupported_Float_Width
+              with "Unsupported float width: " &  Integer'Image (Float_Width);
+      end case;
+   end To_Float_Format;
+
+   function Float_Mantissa_Size (Float_Type : Irep) return Integer
+   is
+      Format : constant Float_Format := To_Float_Format (Float_Type);
+   begin
+      case Format is
+         when IEEE_32_Bit =>
+            return 23;
+         when IEEE_64_Bit =>
+            return 52;
+      end case;
+   end Float_Mantissa_Size;
+
    ---------------------
    -- Name_Has_Prefix --
    ---------------------
--- a/gnat2goto/driver/range_check.adb
+++ b/gnat2goto/driver/range_check.adb
@@ -1,4 +1,6 @@
 with Binary_To_Hex;         use Binary_To_Hex;
+with Types;                 use Types;
+with GOTO_Utils;             use GOTO_Utils;
 
 package body Range_Check is
 
@@ -14,12 +16,26 @@ package body Range_Check is
       return Length;
    end Store_Bound;
 
+   ----------------------
+   -- Store_Real_Bound --
+   ----------------------
+
+   function Store_Real_Bound (Number : Bound_Type_Real) return Integer
+   is
+      Length : constant Integer := Integer (Integer_Bounds_Real_Table.Length);
+   begin
+      Integer_Bounds_Real_Table.Append (Number);
+      return Length;
+   end Store_Real_Bound;
+
    -----------------------
    -- Load_Bound_In_Hex --
    -----------------------
 
-   function Load_Bound_In_Hex (Index : Integer; Bit_Width : Pos) return String
+   function Load_Bound_In_Hex (Index : Integer; Actual_Type : Irep)
+                               return String
    is
+      Bit_Width : constant Pos := Pos (Get_Width (Actual_Type));
       Bound : constant Uint := Uint (Integer_Bounds_Table.Element (Index));
    begin
       return Convert_Uint_To_Hex (
@@ -27,4 +43,20 @@ package body Range_Check is
                  Bit_Width => Bit_Width);
    end Load_Bound_In_Hex;
 
+   ----------------------------
+   -- Load_Real_Bound_In_Hex --
+   ----------------------------
+
+   function Load_Real_Bound_In_Hex (Index : Integer; Actual_Type : Irep)
+                                    return String
+   is
+      Bit_Width : constant Float_Format := To_Float_Format (Actual_Type);
+      Bound : constant Ureal :=
+        Ureal (Integer_Bounds_Real_Table.Element (Index));
+   begin
+      case Bit_Width is
+         when IEEE_32_Bit => return Convert_Ureal_To_Hex_32bits_IEEE (Bound);
+         when IEEE_64_Bit => return Convert_Ureal_To_Hex_64bits_IEEE (Bound);
+      end case;
+   end Load_Real_Bound_In_Hex;
 end Range_Check;
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -271,6 +271,12 @@ package body Tree_Walk is
         Post => Kind (Do_Signed_Integer_Definition'Result) =
                   I_Bounded_Signedbv_Type;
 
+   function Do_Floating_Point_Definition (N : Node_Id) return Irep
+     with Pre  => (Nkind (N) = N_Floating_Point_Definition
+                   and then Is_Type (Defining_Entity (Parent (N)))),
+     Post => Kind (Do_Floating_Point_Definition'Result) in
+     I_Floatbv_Type | I_Bounded_Floatbv_Type;
+
    function Do_Simple_Return_Statement (N : Node_Id) return Irep
    with Pre  => Nkind (N) = N_Simple_Return_Statement,
         Post => Kind (Do_Simple_Return_Statement'Result) = I_Code_Return;
@@ -1767,6 +1773,8 @@ package body Tree_Walk is
    function Do_Real_Constant (N : Node_Id) return Irep is
       Ret           : constant Irep := New_Irep (I_Constant_Expr);
       Real_Constant_Type : constant Irep := Do_Type_Reference (Etype (N));
+      Bit_Width : constant Nat := UI_To_Int (Esize (Etype (N)));
+      Unknown_Float_Width : exception;
    begin
       Set_Source_Location (Ret, Sloc (N));
       Set_Type (Ret, Real_Constant_Type);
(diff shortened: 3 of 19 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -2094,17 +2101,28 @@ package body Tree_Walk is
...
             Set_Rhs (Op_Geq, Adjusted_Lower_Bound);
             Set_Type (Op_Geq, Make_Bool_Type);
             Set_Lhs (Op_Leq, Adjusted_Value_Expr);
             Set_Type (I     => Upper_Bound,
                       Value => Bound_Type);
-            Set_Value (I     => Upper_Bound,
-                      Value => Load_Bound_In_Hex (Get_Upper_Bound (Bound_Type),
-                        Pos (Get_Width (Bound_Type))));
+
+            if Kind (Bound_Type) = I_Bounded_Signedbv_Type then
```
hannes-steffenhagen-diffblue: Might be worth having a helper function that does this dispatch for us?

gnat2goto/driver/tree_walk.adb
```diff
@@ -2094,17 +2101,28 @@ package body Tree_Walk is
...
+               Set_Value (Lower_Bound,
+                          Load_Bound_In_Hex (Get_Lower_Bound (Bound_Type),
+                            Pos (Get_Width (Bound_Type))));
+               Set_Value (Upper_Bound,
+                          Load_Bound_In_Hex (Get_Upper_Bound (Bound_Type),
+                            Pos (Get_Width (Bound_Type))));
+            else
+               Set_Value (Lower_Bound,
+                          Load_Real_Bound_In_Hex (Get_Lower_Bound (Bound_Type),
+                                                  Get_Width (Bound_Type)));
```
hannes-steffenhagen-diffblue: Why is it `Pos (Get_Width (Bound_Type))` in one case and just `Get_Width (Bound_Type)` in the other?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4095,6 +4113,61 @@ package body Tree_Walk is
...
+        New_Irep (if Nkind (Scalar_Range_Ent) = N_Real_Range_Specification
+                  then I_Bounded_Floatbv_Type
+                  else I_Floatbv_Type);
+      Esize_Width : constant Nat := UI_To_Int (Esize (E));
+      Digits_Expr : constant Node_Id := Digits_Expression (N);
+      Digits_Count : constant Uint := Intval (Digits_Expr);
+   begin
+      --  The digit precision is not used yet: this test is here only to show
+      --  how to access it.
+      if Digits_Count > 100 then
```
hannes-steffenhagen-diffblue: I'd remove this before merging, doesn't seem to actually have a function. For what it's worth I don't think we need to care too much about the precision `digits`; The GNAT frontend will have picked a representation for us, and so for us it's just a matter of mapping that representation to our float types.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4095,6 +4113,61 @@ package body Tree_Walk is
...
+      Set_Width (Ret, Integer (Esize_Width));
+      Set_F (Ret,
+             (case Esize_Width is
+                 when 32     => 23,
+                 --  23-bit mantissa, 8-bit exponent
+
+                 when 64     => 52,
+                 --  52-bit mantissa, 11-bit exponent
+
+                 when others => raise Program_Error));
```
hannes-steffenhagen-diffblue: Should have a proper error message rather than a `Program_Error`?

gnat2goto/unit/floating_point_tests.adb
```diff
@@ -34,7 +34,8 @@ package body Floating_Point_Tests is
         & "00000000"
         & "0000000";
    begin
-      pragma Assert (Convert_Ureal_To_Binary_IEEE (Eight) = IEEE_Bits);
+      pragma Assert (Convert_Ureal_To_Binary_IEEE (Eight, 23, 8, 127)
```
hannes-steffenhagen-diffblue: Should use the 32 bit helper rather than repeating the hardcoded values here

xbauch: There are no helpers for the conversion to binary, because I though people wouldn't want to use them directly, i.e. outside the conversion to Hex. I can add the helpers here as well.

gnat2goto/driver/tree_walk.adb
xbauch: Well the int->hex conversion wants bit-width as a `Pos` type. In the real->hex conversion I only use the bit-width in a direct case-analysis, thus I don't care about the type.

gnat2goto/unit/floating_point_tests.adb
hannes-steffenhagen-diffblue: Ah I didn't see that. Hm, in that case your call I'd say :woman_shrugging:

gnat2goto/driver/tree_walk.adb

gnat2goto/driver/tree_walk.adb

gnat2goto/driver/tree_walk.adb

gnat2goto/driver/goto_utils.adb
```diff
@@ -329,6 +329,21 @@ package body GOTO_Utils is
...
+   begin
+      if Entity_Esize = 32 then
+         --  23-bit mantissa, 8-bit exponent
+         return 23;
+      end if;
+      if Entity_Esize = 64 then
+         --  52-bit mantissa, 11-bit exponent
+         return 52;
+      end if;
+      raise Unknown_Float_Width;
```
hannes-steffenhagen-diffblue: We should have a proper human readable error in case this ever happens in practice (which it well might)

gnat2goto/driver/range_check.adb
```diff
@@ -27,4 +39,20 @@ package body Range_Check is
...
+   function Load_Real_Bound_In_Hex (Index : Integer; Bit_Width : Integer)
+                                    return String
+   is
+      Bound : constant Ureal :=
+        Ureal (Integer_Bounds_Real_Table.Element (Index));
+   begin
+      return (case Bit_Width is
+                 when 32 => Convert_Ureal_To_Hex_32bits_IEEE (Bound),
+                 when 64 => Convert_Ureal_To_Hex_64bits_IEEE (Bound),
+                 when others => raise Program_Error);
```
hannes-steffenhagen-diffblue: Maybe just have the 32/64 as a precondition to the function? Or even just use a two value enum like `type Float_Format is (IEEE_32_Bit, IEEE_64_Bit);` together with something like 

```
function To_Float_Format (Float_Width : Integer) return Float_Format is
begin
case Float_Width is
when 32 => return IEEE_32_Bit;
when 64 => return IEEE_64_Bit;
when others =>
  // print some error message along the lines of 'We only support 32 and 64 bit floats'
  raise Unknown_Float_Format;
end To_Float_Format;
```
?

gnat2goto/driver/tree_walk.adb
```diff
@@ -2060,6 +2076,29 @@ package body Tree_Walk is
             Adjusted_Upper_Bound : Irep;
             Source_Location : constant Source_Ptr := Get_Source_Location
               (Value_Expr);
+
+            procedure Dispatch_Bounds;
+
+            procedure Dispatch_Bounds
```
hannes-steffenhagen-diffblue: I meant more something along the lines of

```
function Get_Bound_In_Hex (Bound_Index : Integer) return String is
  Bound_Width : constant Integer := Get_Width (Bound_Type);
begin
return (case Kind (Bound_Type) is
  when I_Bounded_Signedbv_Type => Load_Bound_In_Hex (Bound_Index, Pos (Bound_Width))
  when I_Bounded_Floatbv_Type => Load_Real_Bound_In_Hex (Bound_Index, Bound_Width)));
end Get_Bound_In_Hex;
```

And then later you can just do

```
Set_Value (Lower_Bound, Get_Bound_In_Hex (Get_Lower_Bound (Bound_Type));
Set_Value (Upper_Bound, Get_Bound_In_Hex (Get_Upper_Bound (Bound_Type));
```

gnat2goto/driver/goto_utils.adb
```diff
@@ -329,19 +330,31 @@ package body GOTO_Utils is
...
-      if Entity_Esize = 64 then
-         --  52-bit mantissa, 11-bit exponent
-         return 52;
-      end if;
-      raise Unknown_Float_Width;
+      case Float_Width is
+         when 32 => return IEEE_32_Bit;
+         when 64 => return IEEE_64_Bit;
+         when others =>
+            Raise_Exception (Unsupported_Float_Width'Identity,
```
hannes-steffenhagen-diffblue: Could alternatively be written as `raise Unsupported_Float_Width with "error message";`

gnat2goto/driver/goto_utils.adb
```diff
@@ -329,19 +330,31 @@ package body GOTO_Utils is
...
+                             "Unsupported float width: " &
+                               Integer'Image (Float_Width));
+      end case;
+   end To_Float_Format;
+
+   function Float_Size (Float_Type : Irep) return Integer
```
hannes-steffenhagen-diffblue: Shouldn't this be something like `Float_Mantissa_Size`?

gnat2goto/driver/range_check.adb
```diff
@@ -43,16 +47,17 @@ package body Range_Check is
...
+      Bit_Width : constant Float_Format := To_Float_Format (Actual_Type);
       Bound : constant Ureal :=
         Ureal (Integer_Bounds_Real_Table.Element (Index));
    begin
-      return (case Bit_Width is
-                 when 32 => Convert_Ureal_To_Hex_32bits_IEEE (Bound),
-                 when 64 => Convert_Ureal_To_Hex_64bits_IEEE (Bound),
-                 when others => raise Program_Error);
-   end Load_Real_Bound_In_Hex;
+      if Bit_Width = IEEE_32_Bit then
```
hannes-steffenhagen-diffblue: Would prefer a case statement here, as that would cause a compile time error if we add support for more float types without updating this section

gnat2goto/driver/goto_utils.adb
martin-cs: There is no reason in principle why we couldn't support other widths but it is not a priority for now.

gnat2goto/driver/goto_utils.adb
xbauch: Done.

gnat2goto/driver/range_check.adb
xbauch: Done.

gnat2goto/driver/goto_utils.adb
hannes-steffenhagen-diffblue: Not in principle, no, but we need to decide on a format for encoding the float in - GNAT gives us floating point numbers as arbitrary size rationals we encode ourselves rather than any particular binary representation.

hannes-steffenhagen-diffblue: Although, reading up on it now looks like there's the `Machine_Radix` and `Machine_Mantissa` attributes we could use for this purpose instead. I'll have a look.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4127,27 +4131,11 @@ package body Tree_Walk is
                   then I_Bounded_Floatbv_Type
                   else I_Floatbv_Type);
       Esize_Width : constant Nat := UI_To_Int (Esize (E));
-      Digits_Expr : constant Node_Id := Digits_Expression (N);
-      Digits_Count : constant Uint := Intval (Digits_Expr);
+      --  Digits_Expr : constant Node_Id := Digits_Expression (N);
+      --  Digits_Count : constant Uint := Intval (Digits_Expr);
```
NlightNFotis: Can this commented out code be removed?

PR #150, opened 2019-03-08
Remove warning 'unhandled node N_Renaming_Declaration node'

sonodtt:
We do not need to process the N_Renaming_Declaration node - it is handled transparently by the gnat front-end.

PR also adds another simple example to the relevant test.


hannes-steffenhagen-diffblue:
Had a discussion with @sonodtt and he'd prefer to add a couple more test cases for other uses of rename, especially for packages and so on before merging this, so I'm putting a blocker on it for now to prevent this getting merged by accident

PR #152, opened 2019-03-11
Implement Op_Abs

hannes-steffenhagen-diffblue:
Also changes the abs test slightly to make sure we're testing runtime behavior
rather than the constant propagator.


hannes-steffenhagen-diffblue:
Forgot to commit test.abs, oops

hannes-steffenhagen-diffblue:
@xbauch good point, I'll add a floating point test as well

Issue #149, opened 2019-03-07
floating literals are broken


xbauch:
More specifically the conversion from Ada reals to IEEE float/double (or the subsequent conversion to hex-format) introduces imprecision in the decimal part of the number. For example, 1.1f is converted into 1.8f. Pending #148 the test `floating_point_literals.adb` demonstrates the issue.

martin-cs:
Decimal <-> binary conversion for float is a tricky topic.  I know a
some about it so whoever this is assigned to : if you would like my
help, just ask.

Cheers,
 - Martin

hannes-steffenhagen-diffblue:
@martin-cs If you have some pointers on that I'd be quite grateful. There are probably standard algorithms for that sort of thing but I didn't find them when I looked for them, so I ended up implementing a naive conversion - the current implementation basically always rounds towards 0, which may lead to some unexpected results (this should probably be configurable but I'm not even entirely sure how to implement for instance round to nearest properly).

martin-cs:
So there are a couple of standard algorithms for doing string to real just using machine arithmetic.  They are very detailed ( see https://www.computerworld.com/article/2512252/php-floating-point-bug-crashes-servers.html and the many other versions of this attack ) and I wouldn't recommend them TBH.  If you have access to rationals or other arbitrary precision data types then something like this:

https://github.com/CVC4/CVC4/blob/master/src/util/floatingpoint.cpp#L511

is not particularly efficient but is correct and in that case pretty well tested.  Be careful of off-by-one errors in the bounds.

hannes-steffenhagen-diffblue:
The conversion we need to do is indeed rational -> bitvector, so I'll take a look at that.

PR #154, opened 2019-03-11
Fix floating point conversion

hannes-steffenhagen-diffblue:
This fixes the immediate issue in https://github.com/diffblue/gnat2goto/issues/149

However, this is still a hackjob of an implementation and needs to be redone from the ground up.


xbauch:
@hannes-steffenhagen-diffblue please also mark the floating point test as passing as well.

PR #156, opened 2019-03-12
Add support for N_Block_Statement node

sonodtt:
Support for nested declares. 

NB nested functions are not yet supported but nested declares in nested functions have been considered in this context. 

Shortly to appear: a Github issue w.r. to nested functions - that in the scenario shown in the test-case of this PR, appear to work out of the box (by accessing outer scope variables directly from the symbol table).  We should try a few more scenarios to fully assess support for nested functions.


gnat2goto/driver/tree_walk.adb
```diff
@@ -5110,7 +5124,7 @@ package body Tree_Walk is
             Append_Op (Block, Do_Loop_Statement (N));
 
          when N_Block_Statement =>
-            Warn_Unhandled_Construct (Statement, "block");
+            Append_Op (Block, Do_N_Block_Statement (N));
```
xbauch: Why not append `do_subprogram_or_block` directly?

sonodtt: Interface. Wrapper. Consistent with others.  If impl. changes in future, this does not change.

PR #157, opened 2019-03-13
Bump CBMC submodule

chrisr-diffblue:


chrisr-diffblue:
I think the new assertions came from goto_symex refactoring work.

PR #158, opened 2019-03-14
Add support for character literals.

NlightNFotis:


hannes-steffenhagen-diffblue:
Can you please also add tests for `Wide_Character` and custom character literal support (i.e. `type My_Character is ('a', 'b', 'c', 'd'); X : constant My_Character := 'a';` etc)

gnat2goto/driver/tree_walk.adb
```diff
@@ -1733,6 +1737,23 @@ package body Tree_Walk is
...
    end Do_Constant;
 
+   ---------------------------
+   -- Do_Character_Constant --
+   ---------------------------
+
+   function Do_Character_Constant (N : Node_Id) return Irep is
+      Ret : constant Irep := New_Irep (I_Constant_Expr);
+      Resolved_Type : constant Irep := Do_Type_Reference (Etype (N));
+      Constant_Width : constant Integer := 8;
```
hannes-steffenhagen-diffblue: I do not believe characters are guaranteed to be 8 bit; For example there are `Wide` and `Wide_Wide` characters

hannes-steffenhagen-diffblue: Basically if you're saying we won't support these for now I think that's fine but this should both be documented and made explicit in the code.

NlightNFotis:
@hannes-steffenhagen-diffblue Probably yes for the first question, absolute no for the second one. The second one has nothing to do with the scope of this PR. What you are describing is enumeration types with character literals as enumeration literals. Quoting https://en.wikibooks.org/wiki/Ada_Programming/Types/Enumeration

```
type ABC is ('A', 'B', 'C');
```

> This literal 'A' has nothing in common with the literal 'A' of the predefined type Character (or Wide_Character).

gnat2goto/driver/tree_walk.adb
NlightNFotis: Yes this was an assumption I wanted someone else to check. I will see how to handle this.

hannes-steffenhagen-diffblue:
It has nothing to do with the predefined character _type_. The predefined character _type_ is, however, not special. Character literals are exactly special kinds of enumerations.

NlightNFotis:
@hannes-steffenhagen-diffblue I think this should go in as is, and I'll make a separate PR to support Wide Character.

sonodtt:
Fotis - before you merge please address this:
"Basically if you're saying we won't support these for now I think that's fine but this should both be documented and made explicit in the code."

NlightNFotis:
@sonodtt @hannes-steffenhagen-diffblue Please re-review this. This now supports all Character types, and it adds tests for that.

PR #160, opened 2019-03-15
Tests for 'use type' & 'use all type' clause

sonodtt:
NB our package implementation does not support function calls atm, so 
although 'use type' can work fully if only built-in operators are used, 
any use of functions or overloaded operators or new 'basic operations' 
for a custom type will fail.

### DIFF
--- a/testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
+++ b/testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
@@ -1,8 +1,0 @@
-with External_Types; use type External_Types.New_Integer;
-
-procedure Use_Type_Clause is
-   A : External_Types.New_Integer := 1;
-begin
-   A := A + 1;
-   pragma Assert (A=2);
-end Use_Type_Clause;
--- a/testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/count_types.adb
+++ b/testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/count_types.adb
@@ -1,0 +1,8 @@
+package body Count_Types is
+
+   function Double (X : in out Counter) return Counter is
+   begin
+      return X*2;
+   end Double;
+
+end Count_Types;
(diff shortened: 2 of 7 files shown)


testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
```diff
@@ -1,8 +1,21 @@
...
+
+   B : Count_Types.Counter;
+   C : Count_Types.Counter;
 begin
-   A := A + 1;
-   pragma Assert (A=2);
+   pragma Assert (A=1);
+
+   C := Double(B);  --  Double is accessible due to 'use all type' clause
+   --  next statmente would fail as packages do not support function calls atm
```
hannes-steffenhagen-diffblue: :pick: statement

testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
```diff
@@ -1,8 +1,21 @@
-with External_Types; use type External_Types.New_Integer;
+--  'use type' clause makes visible only the operators associated with a type
+--  (NB but not the type itself, this must still be fully qualified
+with Count_Types; use type Count_Types.Count;
+
+--  'use all type' clause makes visible both the operators and basic operations
```
hannes-steffenhagen-diffblue: :pick: By "basic" operations do you mean [primitive operations](https://en.wikibooks.org/wiki/Ada_Programming/Object_Orientation#Primitive_operations)? That's what they're usually called in Ada related discussions so if that's the case we should use that lingo here as well.

testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
sonodtt: Done.

testsuite/gnat2goto/tests/package_use_type_N_Use_Type_Clause/Use_Type_Clause.adb
sonodtt: Ah yes. Hmm - I think 'basic' would refer to those inherited according to Ada lingo? 
Mixup. Fixed.

sonodtt:
@hannes-steffenhagen-diffblue  re "default field value not working" - not an issue here, but will look into that and related as part of a quick familiarisation with our packages support. More tests for edge cases on the way.

hannes-steffenhagen-diffblue:
@sonodtt The default field thing isn't really related to packages (or rather: I don't think it is). I think it's just something we don't support for records at the moment.

PR #161, opened 2019-03-18
Encapsulation of arrays - and further encapsulation of range checks

sonodtt:
+ CONTENT
- key functions regarding a) arrays have been moved from tree_walk.adb to an
  'Arrays' package and encapsulated (private access to internal utilities)
- additional functions that assist with range checks have been moved to the existing 
  'range_checks' package and additional encapsulation (private access) introduced.

Aside: Some commits are large - but I'm only moving things around. 
Can break these commits down by key function group if individual verification is required. 

+ STATS
- after array cleanup + encapsulation tree_walk is now 4322 lines (from 5322) =>
(-1000) => (-19%)
- after range cleanup + encapsulation tree_walk is now 4111 lines (from 5322) =>
  (-1211) => (-23%)

NB here access privileges from our perspective are - atm - most useful in how
they document at a glance which functions are internal utilities. We're less
interested in formalising client interfaces at this stage in our development.

+ INTENT
The bigger picture is that by placing these functions in their own package:

a) viewing which procedures and data-structures are key to a major feature
becomes much easier: look at the package. 
This should make further development and refactoring easier.

b) the gargantuanness of tree_walk goes down by ~25%.

c) encourage further use of packages and units. 
When we have a clean slate and uncluttered files - its less likely that we'll
value compactness over readability.

+ PROVISOs

Note that I have not moved e.g. the 'packages' work into a separate package - this
is to not change the ground under Trevor's feet - and it is initially easier for
him to work from within tree_walk. 
Much of the package work will be deeply entwined with other functions in tree_walk.


gnat2goto/driver/arrays.adb
```diff
@@ -309,6 +309,14 @@ package body Arrays is
       return Ret;
    end Make_Array_Default_Initialiser;
 
+   -------------------------------------
+   -- Do_Constrained_Array_Definition --
+   -------------------------------------
+
+   --  No difference between representations at the moment:
```
hannes-steffenhagen-diffblue: Has part of this comment gone missing?

gnat2goto/driver/range_check.adb
```diff
@@ -39,8 +45,8 @@ package body Range_Check is
       Bound : constant Uint := Uint (Integer_Bounds_Table.Element (Index));
    begin
       return Convert_Uint_To_Hex (
-                 Value     => Bound,
-                 Bit_Width => Bit_Width);
+                                  Value     => Bound,
```
hannes-steffenhagen-diffblue: stray formatting change?

gnat2goto/driver/arrays.ads
```diff
@@ -26,25 +30,29 @@ package Arrays is
...
    function Get_Array_Component_Type (N : Node_Id) return Entity_Id
      with Post => Is_Type (Get_Array_Component_Type'Result);
 
+   function Do_Slice (N : Node_Id) return Irep
+     with Pre  => Nkind (N) = N_Slice;
+
+   function Do_Indexed_Component (N : Node_Id) return Irep
+     with Pre  => Nkind (N) = N_Indexed_Component;
+
+private
```
hannes-steffenhagen-diffblue: If these functions are 'private', there's technically no need to mention them here in the `.ads` at all, unless you need them in a precondition/postcondition or something like that

sonodtt: Intentionally there as a form of documentation.

gnat2goto/driver/arrays.adb
sonodtt: Nope. That's all there was.

gnat2goto/driver/range_check.adb
sonodtt: Fixed.

PR #165, opened 2019-03-19
Fix large real literals

xbauch:
The buffer for the integer part was not large enough. Test case was updated and
extended.

### DIFF
--- a/gnat2goto/driver/ureal_to_binary.adb
+++ b/gnat2goto/driver/ureal_to_binary.adb
@@ -29,7 +29,7 @@ package body Ureal_To_Binary is
                                               Frac_Part : out String;
                                               Int_End : out Integer;
                                               Frac_End : out Integer)
-     with Pre => (Denominator (Number) > Uint_0);
+     with Pre => (Norm_Den (Number) > Uint_0);
 
    procedure Get_Integer_And_Fractional_Parts (Number : Ureal;
                                                Int_Part : out String;
@@ -38,6 +38,9 @@ package body Ureal_To_Binary is
                                                Frac_End : out Integer) is
       --   Current numerator, we change this throughout the algorithm
       Num : Uint := Norm_Num (Number);
+      --  Norm_Den is what you'd normally think of in a rational number, and
+      --  Denominator is a part of GNATs weird "sometimes rational number,
+      --  sometimes weird exponent notation" thing
       Den : constant Uint := Norm_Den (Number);
 
       function Get_Next_Exponent return Int;
@@ -151,7 +154,9 @@ package body Ureal_To_Binary is
                                          Exponent_Bias : Positive)
                                          return String is
       Is_Negative : constant Boolean := Number < Ureal_0;
-      Int_Part : String (1 .. 2**(Exponent_Bits - 1));
+      --  this buffer needs to be big enough to store the integer part
+      --  which may be one bit wider than the exponent.
+      Int_Part : String (1 .. 2**(Exponent_Bits + 1));
       Frac_Part : String (1 .. 2**(Exponent_Bits - 1));
       Abs_Number : constant Ureal := (if Is_Negative then -Number else Number);
       Result : String (1 .. Fraction_Bits + Exponent_Bits + 1)
@@ -193,10 +198,15 @@ package body Ureal_To_Binary is
                      First_One := False;
                   end if;
                else
-                  Result (Fraction_Bit) := Bit;
-                  Fraction_Bit := Fraction_Bit + 1;
                   Exponent := Exponent + 1;
+                  --  check that we do not start writing past the fraction part
+                  --  (it also marks the end of the result buffer)
+                  if Fraction_Bit <= Fraction_End_Index then
+                     Result (Fraction_Bit) := Bit;
+                     Fraction_Bit := Fraction_Bit + 1;
+                  end if;
                end if;
+
             end loop;
          end Do_Integer_Part;
 
@@ -267,9 +277,6 @@ package body Ureal_To_Binary is
                                       Frac_Part,
                                       Int_End,
                                       Frac_End);
-      if Int_End > Fraction_Bits then
-         raise Integer_Part_Too_Large;
-      end if;
 
       Do_Fraction;
       Exponent := Exponent + Exponent_Bias;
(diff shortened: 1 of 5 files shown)


gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -29,7 +29,7 @@ package body Ureal_To_Binary is
                                               Frac_Part : out String;
                                               Int_End : out Integer;
                                               Frac_End : out Integer)
-     with Pre => (Denominator (Number) > Uint_0);
+     with Pre => (Norm_Den (Number) > Uint_0);
```
hannes-steffenhagen-diffblue: Difference between `Denominator` and `Norm_Den` is that `Norm_Den` is what you'd normally think of in a rational number, and `Denominator` is a part of GNATs weird "sometimes rational number, sometimes weird exponent notation" thing

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -151,7 +145,9 @@ package body Ureal_To_Binary is
                                          Exponent_Bias : Positive)
                                          return String is
       Is_Negative : constant Boolean := Number < Ureal_0;
-      Int_Part : String (1 .. 2**(Exponent_Bits - 1));
+      --  this buffer needs to be big enough to store the integer part
+      --  which is one bit bigger than 2^exp_bits
+      Int_Part : String (1 .. 2**(Exponent_Bits + 1));
```
chrisr-diffblue: A better comment description here might be helpful - the current comment superficially looks to be inconsistent with the code (it's not - it just takes a bit of thinking about to realise this...). A better comment might be something like `...which is one bit wider than the exponent`

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -71,17 +71,11 @@ package body Ureal_To_Binary is
          --  writing to the first position
          Int_End := Int_Part'First + Integer (E);
 
-         if Int_End > Int_Part'Last then
-            raise Integer_Part_Too_Large;
-         end if;
-
-         pragma Assert (Int_End <= Int_Part'Last);
```
hannes-steffenhagen-diffblue: I don't think these bits are necessary anymore with the changes to the buffer size, can you double check?

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -71,17 +71,11 @@ package body Ureal_To_Binary is
...
          --  use the definition of the binary representation to get the integer
          --  part of the fraction, i.e. our integer part is going to be b_E
          --  b_(E-1) ... b_0 (binary digits) with the integer part of
          --  our number being sum (n <- 0..E) { b_n * 2^n}
-         while E >= Uint_0 loop
-            pragma Assert (Digit_Position <= Int_End);
+         while E >= Uint_0 and then Digit_Position <= Int_End loop
```
hannes-steffenhagen-diffblue: Same here

gnat2goto/driver/ureal_to_binary.adb
```diff
@@ -193,10 +189,14 @@ package body Ureal_To_Binary is
                      First_One := False;
                   end if;
                else
-                  Result (Fraction_Bit) := Bit;
-                  Fraction_Bit := Fraction_Bit + 1;
                   Exponent := Exponent + 1;
+                  --  check that we do not start writing past the fraction part
```
chrisr-diffblue: Would be good to explain the implication/result if you *would* have written past the fraction part

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: It's not one bit wider than the exponent though, an 8 bit exponent would have a 512 bit `Int_Part` according to this

xbauch: Done.

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: The fraction bits are what we're trying to write here, so even if the fraction bit wasn't the end of the string we're writing (which it is), it'd still be an unsound thing to do

gnat2goto/driver/ureal_to_binary.adb
chrisr-diffblue: Ah right - so this surely means then that the comment and the code don't match? The code is not `(2^exp_bits)+1` it's `2^(exp_bits+1)` ?

hannes-steffenhagen-diffblue: Yes,  I think the comment is still wrong

gnat2goto/driver/ureal_to_binary.adb
xbauch: May be useful, added as a comment.

gnat2goto/driver/ureal_to_binary.adb
xbauch: Well we need to return `int_end` so the assignment above is necessary.

gnat2goto/driver/ureal_to_binary.adb
xbauch: The check for `digit_position <= int_end` is redundant but it might be nice to keep as documentation.

gnat2goto/driver/ureal_to_binary.adb
xbauch: Done.

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: No, I mean I don't think removing these parts is necessary

gnat2goto/driver/ureal_to_binary.adb
hannes-steffenhagen-diffblue: No, I mean I don't think you need this change, this assertion should always pass now

gnat2goto/driver/ureal_to_binary.adb
xbauch: I see, un-removed. Bellow as well.

gnat2goto/driver/ureal_to_binary.adb
xbauch: Maybe fixed.

peterschrammel:
I was misreading the PR title as `Fix`[`large cardinals`](https://en.wikipedia.org/wiki/Large_cardinal)...

PR #166, opened 2019-03-19
Update tests - checking which tests now pass

sonodtt:


NlightNFotis:
@sonodtt Yes, this is what I tried to do (at a smaller scale albeit), with renaming the operator tests to be prefixed with `op_*`

sonodtt:
@NlightNFotis  I have raised an issue so we don't forget:
https://github.com/diffblue/gnat2goto/issues/168

PR #167, opened 2019-03-19
Array attributes first and last

xbauch:
As far as I could find there are only 4 array attributes and they all work for static arrays (at least to the extend of the added test). For dynamic arrays, this PR adds support for `first` and `last` (length worked before). `Range` for dynamic is still unsupported.

 I also include test for `in_expression` that uses range-attribute but fails since we do not support `n_in`.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -1,4 +1,4 @@
-Occurs: 2448 times
+Occurs: 2447 times
 Calling function: Process_Declaration
 Error message: Pragmas in declaration
 Nkind: N_Pragma
@@ -13,17 +13,12 @@ Calling function: Process_Declaration
 Error message: Unknown declaration kind
 Nkind: N_Freeze_Generic_Entity
 --
-Occurs: 160 times
+Occurs: 159 times
 Calling function: Process_Declaration
 Error message: Number declaration
 Nkind: N_Number_Declaration
 --
-Occurs: 132 times
-Calling function: Do_Expression
-Error message: First attribute
-Nkind: N_Attribute_Reference
---
-Occurs: 121 times
+Occurs: 117 times
 Calling function: Do_Function_Call
 Error message: func name not in symbol table
 Nkind: N_Function_Call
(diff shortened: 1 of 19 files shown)


testsuite/gnat2goto/tests/arrays_attributes/arrays_attributes.adb
```diff
@@ -0,0 +1,17 @@
+procedure Arrays_Attributes is
```
sonodtt: would be helpful to list the attributes at the top in a comment so people know at a glance what to look for:
e.g 
--  array attributes are: first, last, range, length

hannes-steffenhagen-diffblue:
Can you also add a test for dynamic attributes, i.e.

```
procedure Test is 
  type Indet_Array is array(Integer range <>) of Integer;

  procedure Array_Consumer(Arr : Indet_Array) is 
  begin
    pragma Assert(Arr'First = 1);
  end Array_Consumer;

  My_Arr : Indet_Array(1..2) := (others => 0);
begin
  Array_Consumer(My_Arr);
end Test;
```

?

testsuite/gnat2goto/tests/arrays_attributes/arrays_attributes.adb
```diff
@@ -0,0 +1,17 @@
...
+   subtype Range1 is Integer range 10 .. 20;
+   type Arr3 is array (Range1) of Integer;
+   Actual3 : Arr3 := (7, 8, 9, others => 10);
+   subtype My_Index is Integer range Actual3'Range;
+   Array_Attribs : Integer;
+   Val : Integer := 10;
+   An_Index : My_Index := 12;
+begin
+   Actual3 (10 .. 11) := Actual3 (17 .. 18);
+   Actual3 (11) := 100;
```
sonodtt: After 'Actual3 (10 .. 11) := Actual3 (17 .. 18);'
why not a good old pragma 
Actual3(10)=10;

testsuite/gnat2goto/tests/arrays_attributes/arrays_attributes.adb
```diff
@@ -0,0 +1,17 @@
...
+   Actual3 (13 .. 16) := Actual3 (11 .. 12) & Actual3 (15 .. 16);
+   Array_Attribs := Actual3'First + Actual3'Last + Actual3'Length;
+
+   pragma Assert (Array_Attribs = 41);
+   pragma Assert (An_Index > 7);
```
sonodtt: An_index is not reassigned to so seeing this check not as =12 made me check twice if it was assigned to somewhere I'd missed.

testsuite/gnat2goto/tests/arrays_attributes/arrays_attributes.adb
xbauch: Done.

testsuite/gnat2goto/tests/arrays_attributes/arrays_attributes.adb
xbauch: Done.

PR #171, opened 2019-03-22
Implement support for case statements in Ada

NlightNFotis:
This implements support for case statements in Ada. There was already support in for case expressions, which are pretty similar but their evaluation returns a value by default. I heavily modified the support for that to get this to work.

This is still a WIP. I want to add tests (I have them already, but they need to be cleaned up), and I also want to add tests for the case expression feature which seemed to be undocumented and untested as a feature. This is just to facilitate some early reviewing for the actual functionality.


hannes-steffenhagen-diffblue:
Does this work for `when others`?

xbauch:
This looks sensible to me.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1764,6 +1768,57 @@ package body Tree_Walk is
...
+      begin
+         Set_Lhs (Ret, Value);
+         Set_Rhs (Ret, Rhs);
+         Set_Type (Ret, New_Irep (I_Bool_Type));
+         return Ret;
+      end Make_Single_Test;
+
+      This_Alt : Node_Id := First (Alternatives (N));
+   begin
+      --  Do-while loop because there must be at least one alternative.
```
martin-cs: Maybe I am just old and paranoid BUT ... if you are assuming something it is best to make it an assert rather than a comment; even if it is 'obvious'.

hannes-steffenhagen-diffblue: Definitely worth asserting, but the grammar does indeed guarantee this is true for now (but again, worth checking anyway, who knows what's going to happen in Ada 2023)

NlightNFotis:
Just a heads up for this. It covers everything and works reliably for everything **but ranges** in the case statements. That fails, but it fails because `Do_Expression` doesn't handle ranges (this is my understanding at least, because it doesn't recognise the `Node_Irep`, it's falling in the `when others` case).

I am working on it, but it seems to be slightly expanded as scope, so I'd rather this go in as it is (it works and has tests) and then the ranges support goes in as part of a different PR.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1764,6 +1768,79 @@ package body Tree_Walk is
...
+               --  or the last case of complete coverage
+               This_Test := This_Stmt;
+               Append_Op (Ret, This_Test);
+            else
+               This_Test := New_Irep (I_Code_Ifthenelse);
+               Set_Cond (This_Test,
+                         Make_Case_Test
+                           (Discrete_Choices (This_Alt_Copy)));
+               Set_Then_Case (This_Test, This_Stmt);
+               Append_Op (Ret, This_Test);
```
xbauch: I guess this line (and 1829) could be moved after the if-then-else block.

xbauch:
@NlightNFotis I would tend to agree: leave the ranges for another PR and just add a failing test here.

gnat2goto/driver/tree_walk.adb
sonodtt: Parent link of one Hannes posted (case statements)
https://www.adaic.org/resources/add_content/standards/05rm/html/RM-5-4.html

testsuite/gnat2goto/tests/case_statement_vals/case_statement_vals.adb
```diff
@@ -0,0 +1,9 @@
+function Case_Statement_Vals return String is
+  I : Integer := 3;
+begin
+  case I is
+    when 0 | 1 | 2 => return "Valid ternary";
+    when 3         => return "Invalid ternary";
+    when others    => return "";
+  end case;
```
hannes-steffenhagen-diffblue: why does this test not assert this function returns the right value?

testsuite/gnat2goto/tests/case_statement/case_statement.adb
```diff
@@ -0,0 +1,11 @@
+function Case_Statement return String is
+  type Sensor_Type is (Elevation, Azimuth, Distance);
+  Sensor : Sensor_Type := Elevation;
+begin
+  case Sensor is
+    when Elevation => return "Elevation sensor";
+    when Azimuth   => return "Azimuth sensor";
+    when Distance  => return "Distance sensor";
+    when others    => return "";
```
hannes-steffenhagen-diffblue: assert result

testsuite/gnat2goto/tests/case_statement_others_only/case_statement_others_only.adb
```diff
@@ -0,0 +1,8 @@
+function Case_Statement_Others_Only return String is
+  type Sensor_Type is (Elevation, Azimuth, Distance);
+  Sensor : Sensor_Type := Elevation;
+begin
+  case Sensor is
+    when others    => return "";
```
hannes-steffenhagen-diffblue: assert

gnat2goto/driver/tree_walk.adb
```diff
@@ -1764,6 +1768,79 @@ package body Tree_Walk is
...
+         end Make_Single_Test;
+         First_Alt_Test : constant Irep := Make_Single_Test (First (Alts));
+         This_Alt : Node_Id := First (Alts);
+      begin
+         Next (This_Alt);
+         if not Present (This_Alt) then
+            return First_Alt_Test;
+         end if;
+         declare
+            Big_Or : constant Irep := New_Irep (I_Op_Or);
```
hannes-steffenhagen-diffblue: Cute

gnat2goto/driver/tree_walk.adb
```diff
@@ -1764,6 +1768,79 @@ package body Tree_Walk is
...
+         begin
+            Append_Op (Big_Or, First_Alt_Test);
+            Set_Type (Big_Or, New_Irep (I_Bool_Type));
```
hannes-steffenhagen-diffblue: `Make_Bool_Type`, although even better we should probably just have a constant `Bool_Type := Make_Bool_Type` somewhere in this module so we don't keep recreating new ireps for all of these unnecessarily

PR #173, opened 2019-03-22
Further improvements to list_unsupported.sh

chrisr-diffblue:
* Summarization/collation of compilation failures
* Improvements to redacting
* Cleanup to make the above possible


chrisr-diffblue:
Doing this, I've finally got my head wrapped properly around awk :-)

PR #176, opened 2019-03-27
tests for newly found array declaration failure modes

sonodtt:
Key purpose of this PR is to document some ways of declaring multidimensional arrays that causes errors. This also happens for arrays (#177 ). 
The test cases (overall) show what _does_ work and outline (potentially) independent failure modes in separate tests - for clarity and ease of debugging.


testsuite/gnat2goto/tests/arrays_multidimensional_type_instantiation_failure/test.opt
```diff
@@ -0,0 +1 @@
+ALL XFAIL CBMC fails with invariant failure "Condition: op.operands()[number].type() == expr.type()"
```
hannes-steffenhagen-diffblue: missing newline

sonodtt: I thought I'd recently set emacs to save with newline? Grrrrrr...

PR #177, opened 2019-03-27
tests for newly found array declaration failure modes

sonodtt:
Key purpose of this PR is to document a way of declaring arrays that causes errors. This also happens for multidimensional arrays (#176).
Potentially independent failure modes are in separate tests - for clarity and ease of debugging.


hannes-steffenhagen-diffblue:
Is this the same as #176? If not, what's the difference?

sonodtt:
@ Hannes 
Clarified:
Key purpose of this PR is to document a way of declaring arrays that causes errors. This also happens for multidimensional arrays (#176).

Same on twin PR.

PR #179, opened 2019-03-28
Add tests for case expressions

NlightNFotis:
During the work performed for case statement support in gnat2goto (#171) it was discovered that we do support case expressions already, but it's not a documented nor tested feature. This adds some basic tests for this, and fixes any problems discovered.


NlightNFotis:
This is now ready for review. Works reliably, but has the same limitation as #171, in that the cases where there's a range in the condition don't work. This has been described better in #181 .

testsuite/gnat2goto/tests/case_expression_others/case_expression_others.adb
```diff
@@ -0,0 +1,13 @@
+procedure Case_Expression_Others is
+  Invalid_Ternary : Integer := 3;
+  Valid_Result : Integer;
+begin
+  Valid_Result := (case Invalid_Ternary is
+    when 0 => 0,
+    when 1 => 0,
+    when 2 => 0,
+    when others => 1);
```
hannes-steffenhagen-diffblue: same

testsuite/gnat2goto/tests/case_expression_vals/test.opt
```diff
@@ -0,0 +1 @@
+ALL XFAIL CBMC crashes with first argument of if must be boolean, but got or
```
hannes-steffenhagen-diffblue: can you put this together with the commit fixing this so you can remove it?

testsuite/gnat2goto/tests/case_expression_range/case_expression_range.adb
```diff
@@ -0,0 +1,12 @@
+procedure Case_Expression_Range is
+  Random_Val : Integer := 10;
+  Result : Integer;
+begin
+  Result := (case Random_Val is
+    when 1..10 => 1,
+    when 11..20 => 2,
```
hannes-steffenhagen-diffblue: I suspect handling a range expression should be fairly easy - a literal range should always be translatable like this `x in a..b ===> a <= x and x <= b` I'd think

PR #185, opened 2019-03-28
Static constants

tjj2017:
Replace static constants by their static initialisation expression

Please review and merge 
incomplete_type_declaration,
private_type_declaration,
deferred_constant_declaration first.
Then this implementation should be ready for merging.


xbauch:
Was included in #186.

PR #189, opened 2019-04-02
Attributes for discrete types

xbauch:
Pos, Val, Pred, and Succ. Works for discrete, fails for continuous. Succeeding and failing tests are included.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -343,6 +343,30 @@ package body Tree_Walk is
 
    procedure Remove_Entity_Substitution (E : Entity_Id);
 
+   function Do_Attribute_Pos_Val (N : Node_Id) return Irep
+     with Pre => (Ekind (Etype (N)) in Discrete_Kind
+                 and then Nkind (N) = N_Attribute_Reference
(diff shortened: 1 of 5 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -1235,6 +1306,14 @@ package body Tree_Walk is
                when Attribute_Last   =>
                   return Report_Unhandled_Node_Irep (N, "Do_Expression",
                                                      "Last attribute");
+               when Attribute_Val =>
+                  return Do_Attribute_Pos_Val (N);
+               when Attribute_Pos =>
+                  return Do_Attribute_Pos_Val (N);
+               when Attribute_Pred =>
```
sonodtt: could we not have a check for floating point type here and fail with a warning message?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1198,6 +1222,53 @@ package body Tree_Walk is
       return Ret;
    end Do_Enumeration_Definition;
 
+   function Do_Attribute_Pos_Val (N : Node_Id) return Irep is
```
sonodtt: would be nice to have a comment why attribute pos and val are implemented equivalently.

PR #190, opened 2019-04-02
Make uint to (hex or binary) bitvector conversions work better

hannes-steffenhagen-diffblue:
* Sizes no longer have to be a power of two
* Also refactors the uint_to_binary conversion a lot, it is shorter and should
  also be easier to understand now
* Also works with negative numbers now


hannes-steffenhagen-diffblue:
FYI I don't actually know if the "works with negative integers" is new, I just remember that we had an older version that didn't work with negative numbers and this one definitely does :woman_shrugging:

PR #192, opened 2019-04-02
Add support for mod type with arbitrary boundaries

hannes-steffenhagen-diffblue:
This allows us to represent modular types with non-power-of 2
boundaries.

Note: No actual runtime support for any of the operators yet, we will
do this by adding our own implementations and rewriting ada-mod-type to
unsignedbv in a postprocessing step before writing the symbol table.

Also note:

I have no idea how or why the tests pass, considering this isn't actually
replacing the type with something CBMC could be reasonable expected to handle
yet, but I'll take it.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -104,11 +104,6 @@ Error message: Unknown expression kind
 Nkind: N_Freeze_Entity
 --
 Occurs: 5 times
-Calling function: Do_Constant
-Error message: Constant Type not in Class_Bitvector_Type
-Nkind: N_Integer_Literal
---
-Occurs: 5 times
 Calling function: Do_Operator_General
 Error message: Concat unsupported
 Nkind: N_Op_Concat
@@ -149,9 +144,9 @@ Error message: Unknown tree node
 Nkind: N_Compilation_Unit
 --
 Occurs: 1 times
-Calling function: Do_Modular_Type_Definition
-Error message: can not handle modular types with non-power-of-2 bounds at the moment
-Nkind: N_Modular_Type_Definition
+Calling function: Do_Constant
+Error message: Constant Type not in Class_Bitvector_Type
+Nkind: N_Integer_Literal
 --
 Occurs: 1 times
 Calling function: Do_Pragma
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4294,11 +4294,12 @@ package body Tree_Walk is
            (I_Subtype => Make_Nil_Type,
             Width => Mod_Max_Binary_Logarithm);
       end if;
-      return Report_Unhandled_Node_Type
-        (N,
-         "Do_Modular_Type_Definition",
-         "can not handle modular types "
-         & "with non-power-of-2 bounds at the moment");
+
+      return Make_Ada_Mod_Type
+        (I_Subtype => Make_Nil_Type,
+         Width => Mod_Max_Binary_Logarithm,
+         Ada_Mod_Max => Convert_Uint_To_Hex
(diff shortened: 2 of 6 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4215,11 +4215,16 @@ package body Tree_Walk is
...
-         "can not handle modular types "
-         & "with non-power-of-2 bounds at the moment");
+
+      return R : Irep do
+         R := Make_Ada_Mod_Type
+               (I_Subtype => Make_Nil_Type,
+                Width => Mod_Max_Binary_Logarithm,
+                Ada_Mod_Max => Convert_Uint_To_Hex
+                 (Mod_Max, Pos (Mod_Max_Binary_Logarithm)));
+         Set_Ada_Mod_Max (R, Convert_Uint_To_Hex
```
xbauch: Why setting it again?

testsuite/gnat2goto/tests/type_modular_non_power_of_two_decl/modular_type.adb
```diff
@@ -0,0 +1,9 @@
+procedure Modular_Type is
+  type Less_Than_Ten is mod 10;
+  procedure Assert_Val (X : Less_Than_Ten) is
+  begin
+    pragma Assert (X = 5);
```
xbauch: So I guess you won't be able to compare it with anything other than mod-10 type?

hannes-steffenhagen-diffblue: No, that'd be rejected by the frontend.

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: Oops

testsuite/gnat2goto/tests/type_modular_non_power_of_two_decl/modular_type.adb
xbauch: Well on the CBMC side the type is bitvector (but with named sub) so it cannot really be compared with normal bitvectors. But, as you say, that's probably intended.

hannes-steffenhagen-diffblue:
@NlightNFotis my best guess is that cbmc ignored the id and instead decides that it's a bitvector type based on the 'width' field, but I didn't actually check. Note that the next PR will involve replacing all reference to ada-mod-type with the equivalent unsignedbv in the symbol table anyway

experiments/golden-results/StratoX-summary.txt
```diff
@@ -149,9 +144,9 @@ Error message: Unknown tree node
 Nkind: N_Compilation_Unit
 --
 Occurs: 1 times
-Calling function: Do_Modular_Type_Definition
-Error message: can not handle modular types with non-power-of-2 bounds at the moment
-Nkind: N_Modular_Type_Definition
+Calling function: Do_Constant
+Error message: Constant Type not in Class_Bitvector_Type
+Nkind: N_Integer_Literal
```
hannes-steffenhagen-diffblue: I honestly have no idea what this is. Seems like we get a literal 0 **as a type** here for some reason? I don't _think_ this is caused by this PR, but I can't be certain.

xbauch: The report comes from `a-tags.adb` where they declare a range type (which we do not support yet). Then they create constant of that type (which the symbol table does not know). And thus the type is not in `Class_Bitvector_Type`. A bit more magic will have to take place to fix it here, but this is not caused by this PR.

hannes-steffenhagen-diffblue: Thanks! I'd seen that file but I couldn't figure out it was related to range types (wait, what do you mean we don't support those? Isn't that just normal integer types?)

PR #193, opened 2019-04-02
Fix broken tests

hannes-steffenhagen-diffblue:
* attributes test didn't include test.py, and one had wrong test.out
* our testsuite didn't catch that, so I also changed it to detect invalid tests

### DIFF
--- a/testsuite/gnat2goto/testsuite.py
+++ b/testsuite/gnat2goto/testsuite.py
@@ -53,7 +53,7 @@ def main():
     if m.options.discs:
         discs += m.options.discs.split(',')
 
-    test_metrics = {'total': len(test_list), 'uok': 0}
+    test_metrics = {'total': len(test_list), 'uok': 0, 'invalid': 0}
 
     # Generate a standard 'collect_result' function...
     generated_collect_result = generate_collect_result(
@@ -63,8 +63,8 @@ def main():
         metrics = test_metrics)
 
     # ... and then wrap that generated 'collect_result' function in something
-    # that will also accumulate 'UOK' test results
-    def collect_result_including_uok(name, process, _job_info):
+    # that will also accumulate 'UOK' test results and failed tests
+    def collect_test_metrics(name, process, _job_info):
         generated_collect_result(name, process, _job_info)
         test_name = os.path.basename(name)
         test_result = split_file(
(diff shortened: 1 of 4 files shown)


testsuite/gnat2goto/testsuite.py
```diff
@@ -74,19 +74,26 @@ def collect_result_including_uok(name, process, _job_info):
...
+    if test_metrics['invalid'] > 0:
+        print 'Some tests are invalid'
 
     # Generate the report file
     ReportDiff(m.options.output_dir,
                m.options.old_output_dir).txt_image(m.options.report_file)
 
-    if test_metrics['failed'] > 0 or test_metrics['crashed'] > 0 or test_metrics['uok'] > 0:
-        sys.exit(1)
+    if test_metrics['failed'] > 0 or \
```
chrisr-diffblue: Do you actually need the backslashes here?

testsuite/gnat2goto/testsuite.py
```diff
@@ -74,19 +74,26 @@ def collect_result_including_uok(name, process, _job_info):
             test_status = test_result[0].split(':')[0]
             if test_status == 'UOK':
                 test_metrics['uok'] += 1
+            elif test_status == 'INVALID_TEST':
```
chrisr-diffblue: Probably worth changing the name of this method and it's comment now?

testsuite/gnat2goto/testsuite.py
```diff
@@ -74,19 +74,26 @@ def collect_result_including_uok(name, process, _job_info):
...
+                test_metrics['invalid'] += 1
 
     run_testcase = generate_run_testcase('run-test', discs, m.options)
 
     MainLoop(test_list, run_testcase, collect_result_including_uok, m.options.mainloop_jobs)
 
     print "Summary: Ran %(run)s/%(total)s tests, with %(failed)s failed, %(crashed)s crashed, %(uok)s unexpectedly passed." % test_metrics
```
chrisr-diffblue: Could you include the 'invalid' count in the Summary line printed here.

testsuite/gnat2goto/testsuite.py
hannes-steffenhagen-diffblue: it didn't work without them, but honestly my python is kind of weak so maybe there's another way?

chrisr-diffblue: https://www.python.org/dev/peps/pep-0008/#multiline-if-statements maybe?

testsuite/gnat2goto/testsuite.py
hannes-steffenhagen-diffblue: Hm, I deliberately excluded it because an invalid test is more of an exception rather than a test result, in my mind they were two different things.

chrisr-diffblue: I personally didn't see that line so much as "test results" and more "one line summary of what happened". Not a big deal either way, but its sometimes nice to have that summary.

PR #194, opened 2019-04-03
Return nil type where type is expected [ADA-273]

xbauch:
Especially do_full_type_declaration didn't like working with a non-type. Also highlights the fact that access types are still unsupported.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -3715,8 +3715,11 @@ package body Tree_Walk is
             return Do_Modular_Type_Definition (N);
          when N_Floating_Point_Definition =>
             return Do_Floating_Point_Definition (N);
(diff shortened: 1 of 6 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -3715,8 +3715,11 @@ package body Tree_Walk is
             return Do_Modular_Type_Definition (N);
          when N_Floating_Point_Definition =>
             return Do_Floating_Point_Definition (N);
+         when N_Access_To_Object_Definition =>
```
hannes-steffenhagen-diffblue: The change makes sense, but it's a bit surprising because neither commit message or body mention anything about access types.

xbauch: Added.

PR #198, opened 2019-04-03
Replace mod type with unsignedbv before writing JSON symtab

hannes-steffenhagen-diffblue:
For some reason CBMC seems to be able to handle the Ada mod type well enough,
but there is no particular reason to keep the ada-specific irep around, we
really only use it to keep track of type information we don't need at 'runtime'.

Note that this is done by just expanding the scope of the existing `Remove_Bounds` function and appropriately renaming it, I figured it'd make more sense to do it this way considering it was doing more or less the same thing.

### DIFF
--- a/gnat2goto/ireps/ireps_generator.py
+++ b/gnat2goto/ireps/ireps_generator.py
@@ -368,7 +368,7 @@ class IrepsGenerator(object):
                 write(b, "Integer (Follow_Irep (Irep (%s), Follow_Symbol));" % tbl_field)
         return needs_null
 
-    def remove_bounds_set_all_subs(self, b, sn, subs, i, needs_null):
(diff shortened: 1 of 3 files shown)


gnat2goto/ireps/ireps_generator.py
```diff
@@ -2054,6 +2055,10 @@ def generate_code(self, optimize, schema_file_names):
         with indent(b):
             write(b, "return Make_Floatbv_Type (Get_Subtype (I), Get_Width (I), Get_F (I));")
         write(b, "end if;")
+        write(b, "if Kind (I) = I_Ada_Mod_Type then")
+        with indent(b):
+            write(b, "return Make_Unsignedbv_Type (Make_Nil_Type, Get_Width (I));")
```
sonodtt: you can use make_signedint_type

PR #205, opened 2019-04-17
Add a test that demonstrates support added for modular types

NlightNFotis:
Add a test that demonstrates support added for modular types. This test doesn't demonstrate support for modular type arithmetic, and it's literally just the code that was demonstrated earlier to be problematic as part of a bug report, so that we can show it's working.


NlightNFotis:
Hannes suggested that it's better if we modify the existing test, rather than adding a new one. I agree, therefore I'll close this PR and make a new one modifying the test that's under `type_modular/`

PR #206, opened 2019-04-17
Add an extra case to the modular type test.

NlightNFotis:
This is #205, reworked after a suggestion from Hannes.

### DIFF
--- a/testsuite/gnat2goto/tests/type_modular/modular.adb
+++ b/testsuite/gnat2goto/tests/type_modular/modular.adb
@@ -6,7 +6,17 @@ procedure Modular is
   type Two_Bit is mod 4;
   function Next (X : Two_Bit) return Two_Bit is (X + 1);
   Number : constant Two_Bit := 3;
+
+  type Unsigned_64 is mod 2**64;
+  function Next (X : Unsigned_64) return Unsigned_64 is (X + 1);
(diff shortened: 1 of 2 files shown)


testsuite/gnat2goto/tests/type_modular/modular.adb
```diff
@@ -6,7 +6,12 @@ procedure Modular is
   type Two_Bit is mod 4;
   function Next (X : Two_Bit) return Two_Bit is (X + 1);
   Number : constant Two_Bit := 3;
+
+  type Unsigned_64 is mod 2**64;
+  Zero_Unsigned : constant Unsigned_64 := 0;
```
hannes-steffenhagen-diffblue: Would you mind renaming these and using a `Next` function like for the other cases for consistencies sake?

NlightNFotis:
Can you take one more look at this @hannes-steffenhagen-diffblue ?

testsuite/gnat2goto/tests/type_modular/modular.adb
NlightNFotis: None of these add any value what so ever. For a `Next` function, I'm not going to use it in any way, and to make it prose `Unsigned_Sixty_Four_Bit_Value` seems annoying.

PR #208, opened 2019-04-24
Feature modular arithmetic expon

hannes-steffenhagen-diffblue:
Adds support for the `**` operator for `mod` types, also some more utility functions for creating integer constants and function calls.

Golden results might be out of date, but I don't understand what exactly causes the differences (they seem unrelated to the new functionality?)

### DIFF
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -442,4 +442,35 @@ package body GOTO_Utils is
       return False;
    end Has_GNAT2goto_Annotation;
 
+   function Integer_Constant_To_Expr
+     (Value : Uint;
+      Expr_Type : Irep;
+      Source_Location : Source_Ptr)
+   return Irep is
+      Value_Hex : constant String := Convert_Uint_To_Hex
+        (Value => Value,
+         Bit_Width => Pos (Get_Width (Expr_Type)));
+   begin
+      return Make_Constant_Expr
+        (Source_Location => Source_Location,
+         I_Type => Expr_Type,
+         Value => Value_Hex);
+   end Integer_Constant_To_Expr;
+
+   function Make_Simple_Side_Effect_Expr_Function_Call
+     (Arguments : Irep_Array;
+      Function_Expr : Irep;
+      Source_Location : Source_Ptr) return Irep is
+      Argument_List : constant Irep := Make_Argument_List;
+   begin
+      for Argument of Arguments loop
+         Append_Argument (Argument_List, Argument);
+      end loop;
+      return Make_Side_Effect_Expr_Function_Call
+        (Arguments => Argument_List,
+         I_Function => Function_Expr,
+         I_Type => Get_Return_Type
+           (Get_Type (Function_Expr)),
+         Source_Location => Source_Location);
+   end Make_Simple_Side_Effect_Expr_Function_Call;
 end GOTO_Utils;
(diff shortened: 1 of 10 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -2806,6 +2809,204 @@ package body Tree_Walk is
...
+        (Source_Location => Source_Location);
+      Multiply_Loop : constant Irep := Make_Code_While
+        (Loop_Body => Multiply_Loop_Body,
+         Cond => Expon_Greater_Zero,
+         Source_Location => Source_Location);
+      If_Even_Or_Odd_Exponent : constant Irep := Make_Code_Ifthenelse
+        (Cond => Expon_Divisible_By_Two,
+         Then_Case => If_Even_Body,
+         Else_Case => If_Odd_Body,
+         Source_Location => Source_Location);
```
hannes-steffenhagen-diffblue: Apologies for the long function, but I genuinely can't think of a way to make this shorter. :frowning_face:

gnat2goto/driver/goto_utils.adb
```diff
@@ -442,4 +442,35 @@ package body GOTO_Utils is
       return False;
    end Has_GNAT2goto_Annotation;
 
+   function Integer_Constant_To_Expr
```
xbauch: This seems similar to `Build_Index_Constant`.

gnat2goto/driver/goto_utils.ads
```diff
@@ -7,6 +7,8 @@ with Uintp;                 use Uintp;
 
 package GOTO_Utils is
 
+   type Irep_Array is array (Integer range <>) of Irep;
```
xbauch: We already have `Irep_Array` in `arrays.ads`

hannes-steffenhagen-diffblue: Indeed. Should I move it? It seems more like a utility rather than an array-specific thing.

gnat2goto/driver/tree_walk.adb
```diff
@@ -2806,6 +2809,204 @@ package body Tree_Walk is
...
+      Expon_Result : constant Irep := Fresh_Var_Symbol_Expr
+        (Ty => Mod_Type,
+         Infix => "mod_expon_result");
+      Declare_Expon_Result : constant Irep := Make_Code_Decl
+        (Symbol => Expon_Result,
+         Source_Location => Source_Location);
+      Expon_Divisible_By_Two : constant Irep := Make_Op_Eq
+        (Lhs => Make_Op_Mod
+           (Lhs => Exponent_Sym,
+            Rhs => Integer_Constant_To_Expr
```
xbauch: This and other integer constants are used multiple times: could be made shorter if you stored and reused their `irep`.

hannes-steffenhagen-diffblue:
Disclaimer: I have no idea how to even interpret the golden result changes.

gnat2goto/driver/goto_utils.adb
hannes-steffenhagen-diffblue: I'll keep this function for now as I'm not convinced `Build_Index_Constant` is doing the right thing by just converting silently converting non-bitvector integers to 32 bit integers

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: I considered it, but I'm not sure if it's enough of an improvement to be worth doing.

hannes-steffenhagen-diffblue:
Disclaimer: I am an idiot, I used the wrong script to update the golden results before. Now they make a whole lot more sense.

PR #209, opened 2019-04-29
Failing tests for private type declarations

xbauch:
There are 3 cases in our golden results where we still do not support private type declarations:

1: abstract type declaration
2: using private type as array type
3: private type extension


testsuite/gnat2goto/tests/private_type_dec_abstract/test.opt
```diff
@@ -0,0 +1 @@
+ALL XFAIL gnat2goto fails with abstract type declaration unsupported
```
hannes-steffenhagen-diffblue: Supporting tagged types is going to be fun if we ever get around to it

PR #210, opened 2019-04-29
Add support for qualified expressions

NlightNFotis:
Add support for qualified expressions. They look like this:

```ada
type MASK is (FIX, DEC, EXP, SIGNIF);
type CODE is (FIX, CLA, DEC, TNZ, SUB); 

PRINT (MASK'(DEC));  --  DEC is of type MASK
PRINT (CODE'(DEC));  --  DEC is of type CODE 
```

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -54,6 +54,9 @@ package body Tree_Walk is
    with Pre => Nkind (N) = N_Case_Expression,
         Post => Kind (Do_Case_Expression'Result) = I_Let_Expr;
 
+   function Do_Qualified_Expression (N : Node_Id) return Irep
+   with Pre => Nkind (N) = N_Qualified_Expression;
+
    function Do_Constant (N : Node_Id) return Irep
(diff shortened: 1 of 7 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -1432,6 +1433,22 @@ package body Tree_Walk is
...
    end Make_Assume_Expr;
 
+   -----------------------------
+   -- Do_Qualified_Expression --
+   -----------------------------
+
+   function Do_Qualified_Expression (N : Node_Id) return Irep is
+      Value : constant Irep := Do_Expression (Expression (N));
+      Type_Of_Val : constant Irep :=
+         Follow_Symbol_Type (
```
xbauch: Is following necessary here?

NlightNFotis: Without it it wasn't getting the correct bounds.

hannes-steffenhagen-diffblue: Without the follow you're getting a symbol type here, and `Make_Range_Assert` is expecting a resolved type instead

PR #212, opened 2019-05-01
Support for function bodies

xbauch:
1: allow calling function from packages.
2: allow linking two subprogram bodies.

Some tests needed to be changed as well as the py script calling them. This won't work until CBMC supports resolving conflicting JSON symtabs (or we stop gnat2goto producing them).


testsuite/gnat2goto/tests/constant_declarations/test.adb
```diff
@@ -0,0 +1,9 @@
+with Const_Decs; use Const_Decs;
+
+procedure Test is
+   My_P : Integer := 4;
+begin
+   Inc (My_P);
+   --  Will fail: symex does not see the value
+   pragma Assert (My_P = 50);
```
hannes-steffenhagen-diffblue: With this one I'm confused. Is this supposed to happen or not? Because I'd think after `Inc (My_P)` `My_P` would be 5, not 50?

testsuite/gnat2goto/tests/deferred_constant/test.adb
```diff
@@ -0,0 +1,7 @@
+with Deferred_Const; use Deferred_Const;
```
hannes-steffenhagen-diffblue: ?

testsuite/gnat2goto/tests/incomplete_dec/test.adb
```diff
@@ -0,0 +1,8 @@
+with Incomplete_Dec; use Incomplete_Dec;
+
+procedure Test is
+   My_Rec : Partial_Dec := (A=>1);
+begin
+   P(My_Rec);
```
hannes-steffenhagen-diffblue: What's `P`? `Inc`?

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -29,21 +29,18 @@ def filter_timing(results):
 
     return re.sub(skip, r'', results)
 
-
-def process(debug, file, cbmcargs):
+def process_gnat2goto(debug, file, not_main):
```
hannes-steffenhagen-diffblue: :pick: I always get confused with 'not_' boolean flags, I'd flip it around and call this `is_main` and then do the negation when you ask for it. Alternatively, just use the name of the option, i.e. `no_cprover_start`.

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -60,12 +57,23 @@ def process(debug, file, cbmcargs):
...
+    for file in files:
+        unit          = os.path.splitext(file)[0]
+        file_jsout    = unit + ".json_symtab"
+        jsout.append(file_jsout)
 
-    # only run the following if gnat2goto succeeded
-    # Run(["cbmc", jsout, "--show-symbol-table"], output=symtab)
-    # Run(["cbmc", jsout, "--show-goto-functions"], output=gotoprog)
-    cmdline = ["cbmc", jsout]
+    out      = unit + ".out"
```
hannes-steffenhagen-diffblue: This is pretty confusing now. You're just using the name of the last file as `out`?

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -91,16 +98,19 @@ def filter_cbmc_output(cbmc_output):
     lines = cbmc_output.split("\n")
     return "\n".join(filter(cbmc_match, lines))
 
-def prove(cbmcargs="", debug=False):
+def prove(cbmcargs="", main="", debug=False):
     """Call gnat2goto (and cbmc) on all *.adb files from the current directory
 
     PARAMETERS
       none: yet
```
hannes-steffenhagen-diffblue: This docstring is lying about the parameters of this function.

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -91,16 +98,19 @@ def filter_cbmc_output(cbmc_output):
...
-            print('<<< DEBUG ' + file + ' >>>')
-            print(out)
-            print('<<< END DEBUG ' + file + ' >>>')
-        print(filter_cbmc_output(out))
+        gnat2goto_success = process_gnat2goto(debug, file, main != "" and file != main)
+        if gnat2goto_success == False:
+            return ""
+    out = process_cbmc(debug, ada_body_files(), cbmcargs)
+    if debug:
+        print('<<< DEBUG ' + file + ' >>>')
```
hannes-steffenhagen-diffblue: Debug output (I'm guessing this'll go away once the draft phase is done)

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -91,16 +98,19 @@ def filter_cbmc_output(cbmc_output):
...
     """
     for file in ada_body_files():
-        out = process(debug, file, cbmcargs)
-        if debug:
-            print('<<< DEBUG ' + file + ' >>>')
-            print(out)
-            print('<<< END DEBUG ' + file + ' >>>')
-        print(filter_cbmc_output(out))
+        gnat2goto_success = process_gnat2goto(debug, file, main != "" and file != main)
+        if gnat2goto_success == False:
```
hannes-steffenhagen-diffblue: why not just `if not process_gnat2goto(...)` instead?

testsuite/gnat2goto/tests/constant_declarations/test.adb
tjj2017: Perhaps the name for Inc is inappropriate. The subprogram actually adds a number based on the variable Var in the constant_decs package.  Using the value that Var is initialised with, Inc adds 46 to its parameter.  I think a more usual scenario that I think we should test is when a private type and deferred constant is used (as in the test below.

testsuite/gnat2goto/tests/deferred_constant/test.adb
tjj2017: I am not sure what this test is checking.  Originally I provide a test program called use_deferred_const

with Deferred_Const; use Deferred_Const;
procedure Use_Deferred_Const is
   No_Use : Priv := Def_Const;
begin
   Inc (No_Use);
end Use_Deferred_Const;

If we add another priv constant to the package Deferred_Const as below:

package Deferred_Const is
   type Priv is private;
   
   Def_Const : constant Priv;
   Def_Const_Plus_1 : constant Priv;
   
   procedure Inc (P : in out Priv);
   
private
   type Priv is range 1 .. 10;
   
   Def_Const : constant Priv := 1;
   Def_Const_Plus_1 := Def_Const + 1;
end Deferred_Const;

Then in Use_Deferred_Constant after the call to Inc we could place the
pragma Assert (No_Use = Def_Const_Plus_1);

Which should succeed.

testsuite/gnat2goto/tests/incomplete_dec/test.adb
tjj2017: P is just an increment routine.  Perhaps better to rename.  In retrospect I need not have made the test for an incomplete type declaration in a package I could have easily written:

procedure New_Incomlete_Dec is
   type Partial_Dec;

   procedure Inc (X : in out Partial_Dec);

   type Partial_Dec is record
      A : Integer;
   end record;

   procedure Inc (X : in out Partial_Dec) is
   begin
      X.A := X.A + 1;
   end Inc;

   V : Partial_Dec := (A => 19);

begin
   Inc (V);
   pragma Assert (V.A = 20);
end New_Incomlete_Dec;

testsuite/gnat2goto/tests/deferred_constant/test.adb
xbauch: I haven't added the test, I just wanted it to reach symex. This way the the range-check for `Def_Const` is included in the verification. Even if we don't actually refer to anything from the package.

testsuite/gnat2goto/tests/constant_declarations/test.adb
xbauch: Renamed to `Add_46`.

testsuite/gnat2goto/tests/incomplete_dec/test.adb
xbauch: Renamed.

testsuite/gnat2goto/lib/python/test_support.py
xbauch: I went for `is_main`.

testsuite/gnat2goto/lib/python/test_support.py
xbauch: I guess I was. What about `TEST_NAME` instead of `unit` here?

testsuite/gnat2goto/lib/python/test_support.py
xbauch: Added.

testsuite/gnat2goto/lib/python/test_support.py
xbauch: I did not add it, but I can remove it. Done.

testsuite/gnat2goto/lib/python/test_support.py
xbauch: So that's the syntax, ok.

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -102,15 +101,13 @@ def prove(cbmcargs="", main="", debug=False):
...
+      debug: flag indicating if debug output should be printed out
     """
     for file in ada_body_files():
-        gnat2goto_success = process_gnat2goto(debug, file, main != "" and file != main)
-        if gnat2goto_success == False:
+        gnat2goto_success = process_gnat2goto(debug, file, main == "" or file == main)
+        if not gnat2goto_success:
             return ""
     out = process_cbmc(debug, ada_body_files(), cbmcargs)
-    if debug:
```
hannes-steffenhagen-diffblue: Shame on me, I didn't realise this was controlled through a flag. Maybe this shouldn't be removed after all, sorry

hannes-steffenhagen-diffblue:
@xbauch Now that merging symtabs is supported can this be un-drafted?

testsuite/gnat2goto/lib/python/test_support.py
xbauch: Un-removed.

Issue #56, opened 2018-10-30
Refactor ireps_generator.py and add a high level API for generating ada code


hannes-steffenhagen-diffblue:
Right now `ireps_generator` is a 2k+ LOC file with lots of state and write block interspersed everywhere. For maintainability it'd be better to split it up into several files by tasks.

Also, the way we generate the code right now we just have a bunch of write statements everywhere which are a bit hard to follow, I'm thinking if we had something like this (pseudocode):

```
irep_id_arg = ada.variable(name = 'S', type = ada.string_type)
trivial_irep_function = ada.function(name = 'Trivial_Irep', args = [irep_id_arg])
trivial_irep_result = ada.variable(name = 'V', type = json_value_type)
trivial_irep_function.declare(trivial_irep_result)
trivial_irep_function.body([
  v.method_call('Set_Field', ['id', irep_id_arg]),
  v.method_call('Set_Field', ['sub', empty_array.call()]),
  v.method_call('Set_Field', ['namedSub', create_object.call()]),
  v.method_call('Set_Field', ['comment', create_object.call()]),
  ada.return(v)
])
ireps_module.define_private_function(trivial_irep_function)
```

to produce code like this:

```
   function Trivial_Irep (S : String) return JSON_Value is
   begin
      return V : constant JSON_Value := Create_Object do
         V.Set_Field ("id", S);
         V.Set_Field ("sub", Empty_Array);
         V.Set_Field ("namedSub", Create_Object);
         V.Set_Field ("comment", Create_Object);
      end return;
   end Trivial_Irep;
```

Instead of having a bunch of `file.write("Some_Ada_Code")` calls everywhere.

Basically, have a python side Ada AST and print it. I'm not suggesting we do this _now_, but that'll probably be easier to maintain than having random print calls throughout the code everywhere so I'd prefer if we moved to something like that if we find the time to do it.

PR #213, opened 2019-05-02
Derived type tests

tjj2017:
Added some tests to test suite to check the processing of simple derived types (untagged types) in response to issue #56.

Test shows simple derived types are handled but a derived private type causes a gnat2goto failure.  Possibly due to not handling type conversions?

### DIFF
--- a/test.opt
+++ b/test.opt
@@ -1,0 +1,1 @@
+ALL XFAIL gnat2goto fails with private type declarations unsupported
--- a/test.opt
+++ b/test.opt
@@ -1,1 +1,0 @@
-ALL XFAIL gnat2goto fails with private type declarations unsupported
--- a/testsuite/gnat2goto/tests/derived_types/derived_types.adb
+++ b/testsuite/gnat2goto/tests/derived_types/derived_types.adb
@@ -1,0 +1,11 @@
+with Private_Type; use Private_Type;
+procedure Derived_Types is
+   type My_Int is range -23 .. 23;
+   type My_Real is new Float;
+   type Derived_Priv is new Priv;
+begin
+   pragma Assert (0 in My_Int);
+   pragma Assert (11.0 in My_Real);
+   pragma Assert (My_Int'Succ (22) = 23);
+   null;
+end Derived_Types;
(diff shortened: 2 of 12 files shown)


test.opt
```diff
@@ -0,0 +1 @@
+ALL XFAIL gnat2goto fails with private type declarations unsupported
```
xbauch: Where is this file?

testsuite/gnat2goto/tests/derived_types/derived_types.adb
```diff
@@ -0,0 +1,11 @@
+with Private_Type; use Private_Type;
+procedure Derived_Types is
+   type My_Int is range -23 .. 23;
+   type My_Real is new Float;
+   type Derived_Priv is new Priv;
```
xbauch: So here we test that we can declare derived type of a private but cannot use it?

test.opt
NlightNFotis: Yes, this seems to be misplaced?

tjj2017: I think I removed it because we know support private types?

testsuite/gnat2goto/tests/derived_types/derived_types.adb
tjj2017: We should be able to use it but without a call to a subprogram declared in the visible part of package private_type there is not much that we can do.
The assert statement should work but at the moment gnat2goto does not handle it.  I think the failure is something to do with type_conversions.

test.opt
xbauch: It's not whether or not was it removed, but this file is now stored in `gnat2goto/test.opt`, i.e. the top folder. There's no purpose for it there.

chrisr-diffblue: As the other reviewers mentioned, this file should not be in this location. If it's meant to be removed, please remove it from the commit (`git rm`)

tjj2017:
Branch rebased to diffblue/master and corrected

tjj2017:
This PR has been replaced by PR #253.  Which uses the latest diffblue/master

PR #215, opened 2019-05-09
Pragmas in declarations (volatile) [depends-on: #220]

xbauch:
Handle volatile pragmas by updating the symbol in symbol table. Turned out the symbol wasn't there yet. Extended the `Do_Defining_Identifier` to insert the symbol.

The first commit is a squash of the prerequisite PR #220 -- no need to review it here.

### DIFF
--- a/testsuite/gnat2goto/tests/volatile/volatile_type.adb
+++ b/testsuite/gnat2goto/tests/volatile/volatile_type.adb
@@ -1,0 +1,16 @@
+procedure Volatile_Type is
+   Data: Integer;
+   pragma Volatile(Data);
+   Flag: Boolean;
+   pragma Volatile(Flag);
+begin
+   Data := 42;
+   Flag := True;
+
+   if not Flag then
+      Data := 0;
+   end if;
+
+   pragma Assert (Data=42);
+
+end Volatile_Type;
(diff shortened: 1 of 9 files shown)


testsuite/gnat2goto/tests/volatile/volatile_type.adb
```diff
@@ -0,0 +1,16 @@
...
+   pragma Volatile(Flag);
+begin
+   Data := 42;
+   Flag := True;
+
+   if not Flag then
+      Data := 0;
+   end if;
+
+   pragma Assert (Data=42);
```
NlightNFotis: What was the behaviour of symex prior to the previous changes? Would this particular test have failed anyway?

xbauch: No, it would look exactly the same way. Only the symbol table would contain the `volatile` flag and `gnat2goto` would complain about not supporting pragma volatile.

xbauch:
@martin-cs Those changes come from the first commit, where we store the identifier being defined in the symbol table. I created a stand-alone PR for that change: #220.

chrisr-diffblue:
I'll rebase this branch as a new PR and merge. I'll post a new comment here when it's done.

chrisr-diffblue:
The rebased version is in #246.

chrisr-diffblue:
Merged in  #246.

PR #216, opened 2019-05-09
Pragmas in declarations (attribute) [depends-on: #215]

xbauch:
The only attribute in UKNI is `signal` and CBMC seems to ignore it anyway.

The first commit is a squash of prerequisite PR #215 -- no need to review it here.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4683,6 +4683,9 @@ package body Tree_Walk is
 
    procedure Process_Pragma_Declaration (N : Node_Id) is
       procedure Handle_Pragma_Volatile (N : Node_Id);
+      procedure Handle_Pragma_Machine_Attribute (N : Node_Id)
+        with Pre => Nkind (N) in N_Pragma
+        and then Pragma_Name (N) = Name_Machine_Attribute;
 
       procedure Handle_Pragma_Volatile (N : Node_Id) is
          Argument_Associations : constant List_Id :=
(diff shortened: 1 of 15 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4704,6 +4707,32 @@ package body Tree_Walk is
...
+
+         --  second is the attribute as string
+         Second_Argument : constant Node_Id := Next (First_Argument);
+         Attr_String_Id : constant String_Id :=
+           Strval (Expression (Second_Argument));
+         Attr_Length : constant Integer :=
+           Integer (String_Length (Attr_String_Id));
+      begin
+         String_To_Name_Buffer (Attr_String_Id);
+         if Name_Buffer (1 .. Attr_Length) = "signal" then
```
hannes-steffenhagen-diffblue: Might be worth assigning this to a variable if you're using it multiple times (use `renames` if you're worried about making a copy)

testsuite/gnat2goto/tests/pragma_machine_attribute/pragma_machine_attribute.adb
```diff
@@ -0,0 +1,23 @@
...
+      Seconds_Count := Seconds_Count + 1;
+      if Seconds_Count = Interval then
+         Timeout := True;
+      end if;
+   end Timer_Interrupt;
+
+begin
+   Timer_Interrupt;
+
+   pragma Assert (Timeout and then Interval = Seconds_Count);
```
hannes-steffenhagen-diffblue: FWIW we should really use `and` instead of `and then` when we don't explicitly need short-circuiting behaviour.

gnat2goto/driver/tree_walk.adb
xbauch: Renames works as well. Thanks.

testsuite/gnat2goto/tests/pragma_machine_attribute/pragma_machine_attribute.adb
xbauch: Done.

chrisr-diffblue:
I'll rebase this as a new PR.

chrisr-diffblue:
Merged in #247

PR #218, opened 2019-05-09
Handle representation clause size

xbauch:
Only the case where the requested size is smaller than the actual size -- hopefully in that case it should only be an optimisation.


gnat2goto/driver/tree_walk.adb
```diff
@@ -4531,6 +4531,28 @@ package body Tree_Walk is
...
+         Target_Name : constant Irep := Do_Identifier (Name (N));
+         Target_Type : constant Irep :=
+           Follow_Symbol_Type (Get_Type (Target_Name), Global_Symbol_Table);
+      begin
+         if Kind (Target_Type) in Class_Type then
+            if Attr_Id = "size" then
+               if Get_Width (Target_Type) >=
+                 Integer (UI_To_Int (Intval (Expression (N))))
+               then
+                  --  the type is large enough to store values of the size
```
martin-cs: I'm not sure that it is an optimisation.  It matters for things like unchecked conversion.  @tjj2017 can you give some commentary on this?  I have yet to find in the ARM exactly what the rules for `'Size' are.  One key piece of implementation advice from the standard is:

"A Size clause should be supported for an object if the specified Size is at least as large as its
subtype's Size, and corresponds to a size in storage elements that is a multiple of the object's
Alignment (if the Alignment is nonzero)."

experiments/golden-results/UKNI-Information-Barrier-summary.txt
```diff
@@ -107,3 +112,8 @@ Occurs: 1 times
 Redacted compiler error message:
 unmatched actual "REDACTED" in call
 Raw compiler error message:
+--
+Occurs: 1 times
++===========================GNAT BUG DETECTED==============================+
+| GNU Ada (ada2goto) Assert_Failure failed precondition from range_check.ads:29|
```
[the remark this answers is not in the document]
xbauch: It seems that the type with the `size` clause is used in a type conversion later in the program. That is currently unsupported. So it probably only emerged now because we started handling the 'size' clause. We might want to fix the range-check for type conversion in a different PR.

gnat2goto/driver/tree_walk.adb
xbauch: At least for `size` the front-end seems to present the gnat2goto with already altered types. I changed the commit from ignoring to checking that the size specified by the `size` clause is the same as the one the type has in the symbol table. `Component_size` may require more careful handling but I think this is sufficient for `size`.

xbauch:
@martin-cs I tried to address the issues you have raised. What do you think?

gnat2goto/driver/tree_walk.adb
martin-cs: OK; that sounds like a good thing to do.

experiments/golden-results/UKNI-Information-Barrier-summary.txt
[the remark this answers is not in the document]
martin-cs: Yes; definitely do and also in a separate PR.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -18,17 +13,17 @@ Calling function: Process_Pragma_Declaration
...
-Error message: Generic declaration
-Nkind: N_Generic_Subprogram_Declaration
---
 Occurs: 99 times
 Calling function: Do_Function_Call
 Error message: func name not in symbol table
 Nkind: N_Function_Call
 --
-Occurs: 62 times
+Occurs: 99 times
```
martin-cs: Am I right in thinking that these appear because they are the next error after the size issue?

experiments/golden-results/UKNI-Information-Barrier-summary.txt
martin-cs: Yeah... I realise this is out of scope for this PR but can we fix ASAP.  We shouldn't have compile errors in the golden results.

xbauch: We already have a PR for this: https://github.com/diffblue/gnat2goto/pull/225.

experiments/golden-results/StratoX-summary.txt
xbauch: At least from previous experience, we just got a bit further in the AST and found some new unsupported nodes.

PR #219, opened 2019-05-10
Pragma implementation defined

xbauch:
Ignore one more pragma, as per https://github.com/martin-cs/ASVAT/issues/63#issuecomment-490908676


gnat2goto/driver/tree_walk.adb
```diff
@@ -4943,8 +4940,9 @@ package body Tree_Walk is
             --  Ignored for now
               Name_No_Elaboration_Code_All |
             --  Only affects elaboration and linking so can be ignored for now
-              Name_Universal_Aliasing =>
+              Name_Universal_Aliasing |
             --  Optimisation control, should be ignored
+              Name_Implementation_Defined =>
```
hannes-steffenhagen-diffblue: Have a comment here explaining why it's ok to ignore this?

PR #220, opened 2019-05-10
Store defined symbols

xbauch:
during `Do_Defining_Identifier`, in the symbol table ,if they are not present already.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -1207,17 +1207,10 @@ package body Tree_Walk is
       Set_Type            (Sym, Symbol_Type);
 
       if not Global_Symbol_Table.Contains (Sym_Id) then
-         declare
-            Variable_Symbol : Symbol;
-         begin
-            Variable_Symbol.Name       := Sym_Id;
-            Variable_Symbol.BaseName   := Sym_Id;
-            Variable_Symbol.PrettyName := Intern (Get_Name_String (Chars (E)));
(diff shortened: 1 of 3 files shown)


martin-cs:
@tjj2017 : your thoughts on this one would be appreciated.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1151,6 +1152,20 @@ package body Tree_Walk is
       Set_Identifier      (Sym, Unique_Name (E));
       Set_Type            (Sym, Symbol_Type);
 
+      if not Global_Symbol_Table.Contains (Sym_Id) then
```
hannes-steffenhagen-diffblue: Basically, my understanding is this should only happen if we either didn't see the declaration of the type before somehow (which seems unlikely, I hope), or we couldn't parse the declaration (which would explain the many 'fixes' I suppose).

I'm still slightly skeptical this will handle all cases properly, but I'd suspect it's only going to break in cases which are otherwise broken anyway.

hannes-steffenhagen-diffblue:
It'd still be good to have a test case, though if my suspicion is correct it's actually not possible to construct a stable test case for this...

gnat2goto/driver/tree_walk.adb
tjj2017: I tend to agree with this analysis.  I believe it is a general Ada rule that you cannot reference an entity until it has been at least incompletely defined.

chrisr-diffblue:
I'll rebase this PR as a new PR.

PR #221, opened 2019-05-13
Modular subtype

xbauch:
Adds the initial support for modular subtypes. Operators for the new type are nor included in this PR.

### DIFF
--- a/gnat2goto/driver/gnat2goto_itypes.adb
+++ b/gnat2goto/driver/gnat2goto_itypes.adb
@@ -6,6 +6,7 @@ with GOTO_Utils;            use GOTO_Utils;
 with Range_Check;           use Range_Check;
 with Symbol_Table_Info;     use Symbol_Table_Info;
 with Tree_Walk;             use Tree_Walk;
+with Follow;                use Follow;
 
 package body Gnat2goto_Itypes is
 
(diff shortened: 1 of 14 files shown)


experiments/golden-results/vct-summary.txt
```diff
@@ -3,11 +3,6 @@ Calling function: Process_Declaration
 Error message: Exception declaration
 Nkind: N_Exception_Declaration
 --
-Occurs: 61 times
```
martin-cs: Why does this change?

xbauch: That is a bit of a mystery. There is the same thing in #220.

xbauch: Looking at the commit history, it would almost seem that merging 214 and 217 happened at the same time and resulted in a state that would not pass CI.

martin-cs: Is it possible that master or other changes recently merged cause this?

xbauch: Well 214 was about ignoring pragma in declarations.

martin-cs: OK.  Would it be possible to create a separate PR that brings the golden results up to date w.r.t. master?  Having an idea of what issues each patch fixes is really useful from a PM point of view.

xbauch: Sure: https://github.com/diffblue/gnat2goto/pull/222.

martin-cs: Thank you.

gnat2goto/driver/gnat2goto_itypes.adb
```diff
@@ -171,4 +173,56 @@ package body Gnat2goto_Itypes is
...
+        Follow_Symbol_Type (Modular_Type, Global_Symbol_Table);
+
+      S_Range : constant Node_Id := Scalar_Range (N);
+      Lower_Bound : constant Node_Id := Low_Bound (S_Range);
+      Upper_Bound : constant Node_Id := High_Bound (S_Range);
+
+      Lower_Bound_Value : Integer;
+      Upper_Bound_Value : Integer;
+   begin
+
```
hannes-steffenhagen-diffblue: might be worth asserting that `followed_mod_type` is actually an `unsigned` or `mod_type`

PR #223, opened 2019-05-14
Subtype indication can be expanded name as well

xbauch:
so we extend the case analysis.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -1,52 +1,202 @@
-Occurs: 1272 times
+Occurs: 8914 times
 Calling function: Process_Declaration
 Error message: Representation clause declaration
 Nkind: N_Attribute_Definition_Clause
 --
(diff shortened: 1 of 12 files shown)


chrisr-diffblue:
This should be able to run CI without it timing out once https://github.com/diffblue/gnat2goto/pull/229 is merged.

experiments/golden-results/SPARK-tetris-summary.txt
```diff
@@ -13,9 +13,9 @@ Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: Volatile
 Nkind: N_Pragma
 --
-Occurs: 2 times
+Occurs: 1 times
 Calling function: Process_Declaration
-Error message: Representation clause declaration
+Error message: Representation clause unsupported: component_size
```
martin-cs: How is this fix related to the change in this PR?  It seems vaguely like the change in the golden results for the `'Size` patch...

PR #225, opened 2019-05-15
Range check for non-bounded numeric types

xbauch:
Adds support range checking for signed and unsigned bit-vector types. There is a range check as well.

### DIFF
--- a/gnat2goto/driver/range_check.adb
+++ b/gnat2goto/driver/range_check.adb
@@ -72,7 +72,37 @@ package body Range_Check is
       return All_Bounds_Table.Element (Bound_Index);
    end Get_Bound_Type;
 
-   function Get_Bound (Bound_Type : Irep; Pos : Bound_Low_Or_High) return Irep
+   function Get_Bound (N : Node_Id; Bound_Type : Irep; Pos : Bound_Low_Or_High)
+                             return Irep
+   is
+   begin
+      case Kind (Bound_Type) is
+         when I_Bounded_Signedbv_Type
+            | I_Bounded_Floatbv_Type =>
+            return Get_Bound_Of_Bounded_Type (Bound_Type, Pos);
+         when I_Unsignedbv_Type =>
+            declare
+               Type_Width : constant Integer := Get_Width (Bound_Type);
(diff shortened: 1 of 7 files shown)


gnat2goto/driver/range_check.adb
```diff
@@ -72,7 +72,33 @@ package body Range_Check is
...
+      if Kind (Bound_Type) in I_Bounded_Signedbv_Type | I_Bounded_Floatbv_Type
+      then
+         return Get_Bound_Of_Bounded_Type (Bound_Type, Pos);
+      else
+         declare
+            Type_Width : constant Integer := Get_Width (Bound_Type);
+            Bound_Value : constant Uint :=
+              (if Pos = Bound_Low
+               then Uint_0
+               else UI_From_Int (2 ** Type_Width));
```
hannes-steffenhagen-diffblue: `2**Type_Width-1`

hannes-steffenhagen-diffblue:
Can we have tests for this?

martin-cs:
Can we also synchronise this with https://github.com/diffblue/gnat2goto/pull/228

xbauch:
> Can we also synchronise this with #228

Yes, I guess #228 makes the first commit redundant (and insufficient).

xbauch:
@hannes-steffenhagen-diffblue I added the test case (at least of the signed type).

testsuite/gnat2goto/tests/range_check_non_bounded/test.adb
```diff
@@ -0,0 +1,8 @@
+procedure Test is
+   type Uint_8 is mod 2**8;
+   Uint_Var : Uint_8 := 5;
+   Int_Var : Integer;
+begin
+   Int_Var := Integer'(Integer (Uint_Var));
```
hannes-steffenhagen-diffblue: Why both qualified expression _and_ typecast?

xbauch: Qualified expression alone doesn't even parse (expected `integer` type). And typecast alone leads hides the conversion (thus we wouldn't get the range-check).

PR #227, opened 2019-05-16
Check type is known when building reference

xbauch:
Regression tests already exist, e.g. `arrays_access` would fail the test if we did not upgrade the construction of constrained arrays, and `arrays_static_attribute` would fail if we did not declare the implicit type.

### DIFF
--- a/experiments/golden-results/SPARK-tetris-summary.txt
+++ b/experiments/golden-results/SPARK-tetris-summary.txt
@@ -3,6 +3,11 @@ Calling function: Process_Declaration
 Error message: Package declaration
 Nkind: N_Package_Declaration
 --
+Occurs: 11 times
+Calling function: Do_Type_Reference
+Error message: Unsupported declaration of unknown type
+Nkind: N_Defining_Identifier
+--
 Occurs: 10 times
 Calling function: Do_Base_Range_Constraint
 Error message: range expression not bitvector type
--- a/experiments/golden-results/UKNI-Information-Barrier-summary.txt
+++ b/experiments/golden-results/UKNI-Information-Barrier-summary.txt
@@ -3,6 +3,11 @@ Calling function: Process_Declaration
 Error message: Representation clause declaration
 Nkind: N_Attribute_Definition_Clause
 --
+Occurs: 88 times
+Calling function: Do_Type_Reference
+Error message: Unsupported declaration of unknown type
+Nkind: N_Defining_Identifier
+--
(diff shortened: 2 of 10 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4258,7 +4258,21 @@ package body Tree_Walk is
...
    function Do_Type_Reference (E : Entity_Id) return Irep is
-      (Make_Symbol_Type (Identifier => Unique_Name (E)));
+   begin
+      declare
+         Sym_Id : constant Symbol_Id := Intern (Unique_Name (E));
+      begin
+         if not Global_Symbol_Table.Contains (Sym_Id) and Is_Itype (E) then
+            Declare_Itype (E);
+         end if;
+         if not Global_Symbol_Table.Contains (Sym_Id) then
```
martin-cs: If the "checking the same condition twice" is intentional, can we have a comment to stop people refactoring it?  If no, can we refactor it now?

experiments/golden-results/UKNI-Information-Barrier-summary.txt
```diff
@@ -3,6 +3,11 @@ Calling function: Process_Declaration
 Error message: Representation clause declaration
 Nkind: N_Attribute_Definition_Clause
 --
+Occurs: 88 times
```
martin-cs: As this is adding unsupported feature ... was what we were doing before wrong or ... ?

experiments/golden-results/SPARK-tetris-summary.txt
```diff
@@ -3,6 +3,11 @@ Calling function: Process_Declaration
 Error message: Package declaration
 Nkind: N_Package_Declaration
 --
+Occurs: 11 times
+Calling function: Do_Type_Reference
+Error message: Unsupported declaration of unknown type
```
NlightNFotis: I'm confused by this PR. In the first commit, it's supposedly adding the type to the symtab, and then performs a sanity check that ... fails here? Why are these hit? As far as I understood it this shouldn't happen, unless I've misunderstood the code in the first commit. Was the error message supposed to be hit?

xbauch: The first commit only declares implicit types. This report catches the case where the missing type was not implicit.

xbauch:
This was proposing fixing the symptoms, not the cause. We have to find a better solution.

PR #228, opened 2019-05-16
Fix qualified expression

tjj2017:
This is a partial fix preventing gnat2goto aborting from an unsatisfied pre-condition when a qualified expression is not a numeric subtype.
The current changes work for record subtype qualifiers and numeric subtype qualifiers excluding Integer (and possibly Float -  I have not checked this). It does not work for qualified enumeration subtypes.  For type Integer and enumeration subtypes it still fails a pre-condition check.
For qualified array subtypes it does not currently perform a check but should not cause a pre-condition failure.  Eventually a check of the array subtype must be implemented.  At the moment I could not produce a simple test case which got as far as checking an array subtype qualified expression.

The important feature of the partial fix is that it facilitates running list_unsupported script on UKNI without gnat2goto aborting.

I think its shortcomings should be raised as issues.


gnat2goto/driver/tree_walk.adb
```diff
@@ -3495,7 +3493,36 @@ package body Tree_Walk is
...
+      Entity_Node : Node_Id;
+
    begin
+      Def_Unit_Name := Defining_Unit_Name (N);
+
+      --  Defining_Unit_Name will return a N_Defining_Identifier
+      --  for non-child package but a N_Package_Specification when it is a
+      --  child package.
+      --  To obtain the Entity N_Defining_Identifier is required.
+      --  The actual parameter for Unique_Name must  be an Entity node.
```
xbauch: double space

gnat2goto/driver/tree_walk.adb
```diff
@@ -3495,7 +3493,36 @@ package body Tree_Walk is
...
+      if Nkind (Def_Unit_Name) = N_Defining_Identifier then
+         Entity_Node := Def_Unit_Name;
+      else
+         Entity_Node := Defining_Identifier (Def_Unit_Name);
+      end if;
+
+      Package_Name :=
+        Intern (Unique_Name ((Entity_Node)));
```
xbauch: double parentheses

gnat2goto/driver/tree_walk.adb
```diff
@@ -4286,12 +4289,9 @@ package body Tree_Walk is
...
 
    procedure Do_Withed_Unit_Spec (N : Node_Id) is
-      Unit_Name : constant String := Get_Name_String (Get_Unit_Name (N));
    begin
-      if Defining_Entity (N) = Stand.Standard_Standard or else
-        Unit_Name = "system%s"
-      then
-         --  At the moment Standard or System are not processed: TODO
+      if Defining_Entity (N) = Stand.Standard_Standard then
+         --  At the moment Standard is not processed: TODO
```
xbauch: Maybe report as unsupported?

martin-cs:
@chrisr-diffblue and others : with the merged, gnat2goto no longer crashes on UKNI.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1472,15 +1472,28 @@ package body Tree_Walk is
...
+            Type_Of_Val : constant Irep :=
+              Follow_Symbol_Type (Do_Type_Reference (Etype (N)),
+                                  Global_Symbol_Table);
+            Typecast_Expr : constant Irep :=
+              Make_Op_Typecast (Value, Sloc (N), Type_Of_Val);
+         begin
+            return Make_Range_Assert_Expr (N, Typecast_Expr, Type_Of_Val);
+         end;
+      else
+         return Do_Expression (Expression (N));
```
xbauch: We might still want a type-cast here.

gnat2goto/driver/tree_walk.adb
chrisr-diffblue: +1 Yes, please report things that are "TODO" as unsupported features - this might not be important for UKNI, but will be important when gnat2goto is applied to new codebases that we might not have access to. In those cases we really would like to get the most complete list of unsupported features we can.

chrisr-diffblue:
This is currently causing a number of regression/unit tests to fail. These will need to be fixed before we merge this. It may just be that some of the regression tests need their expected output updated.

xbauch:
It's more like it suddenly started processing some standard libraries and is now complaining about pragmas in them (even for regression tests that have no pragmas).

martin-cs:
Yes; looking at the scale of the changes it might be best to split this into multiple PRs with the changes to System first.

gnat2goto/driver/tree_walk.adb
xbauch: This line is also why the regression tests are failing.

martin-cs:
First part is now #230 .

martin-cs:
This PR has now been reduced; other parts in other PRs.

xbauch:
@tjj2017 I have added a report for the non-numeric range check and updated the golden results in #232.

gnat2goto/driver/tree_walk.adb
tjj2017: Removed this mod from this branch.  The branch is now only concerned with qualified expressions

gnat2goto/driver/tree_walk.adb
tjj2017: This mod has now been removed from this branch.  The branch now only deals with qualified expressions

gnat2goto/driver/tree_walk.adb
tjj2017: This mod has now been removed from this branch.  It now only deals with qualified expression.

gnat2goto/driver/tree_walk.adb
tjj2017: It is not really a type cast but a check may be required for array and discriminated types to confirm that the expression is compatible with the given type qualifier.

I have added a "report unhandled node" when array or discriminated qualified expressions are encountered to warn that they are not checked.

tjj2017:
I have added a number of tests for qualified expressions

tjj2017:
Checks for array type qualifiers and discriminated type qualifiers are needed.  There may be some further cases if tagged types are to be handled.

chrisr-diffblue:
@tjj2017 0- Thanks for this work, much appreciated! Rather than doing a git merge of master into this PR, could you instead do a git rebase of this PR - it makes for a much cleaner git history and avoids a few traps if we avoid merge commits in PRs :-)

tjj2017:
This PR has been superseded.  See comment in PR #232

PR #229, opened 2019-05-16
Fix up performance issue in CollectUnsupported

chrisr-diffblue:
I fully expect this to fail CI because I think it will change what the golden results look like - I just want to get a run through CI to see how bad it is, and also get @xbauch views on the code change.

### DIFF
--- a/experiments/collect_unsupported.cpp
+++ b/experiments/collect_unsupported.cpp
@@ -128,7 +128,6 @@ int main(int argc, char** argv)
    multiset<Feature> unsupported_features;
    using Instance = pair<Feature, size_t>;
    vector<Instance> instances;
-   vector<string> params;
 
    string line;
    getline(input_file, line);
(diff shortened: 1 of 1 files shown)


experiments/collect_unsupported.cpp
```diff
@@ -128,25 +128,34 @@ int main(int argc, char** argv)
...
 
    string line;
    getline(input_file, line);
+   unsigned long count=0;
+   cerr << "Scanning " << argv[1] << "...";
    while (true) {
+      ++count;
+      if (count % 1000 == 0) cerr << ".";
       if (line.find("----------At:") == 0) {
+         vector<string> params;
```
xbauch: So each `Feature` was constructed with the parameters of all the previous features? That's unfortunate.

chrisr-diffblue: Yeah, that's what appeared to be happening :-) params.size() was monotonically increasing...

chrisr-diffblue:
Amazingly, this didn't actually change the golden results. Looking closer at the code, I can now see why - and also see why it's such a massive speedup. Each time the params list is used to construct a Feature object, the constructor parses *every* parameter, which basically means that it keeps overwriting fields in the newly constructed object, but since the *last* feature in 'params' is still the one we want, it ends up at the same info - it's just needlessly parsed (and re-parsed, and re-re-parsed, etc) all the rest of the params vector first. :-)

PR #232, opened 2019-05-17
Conditional range check in qualified expression [maybe depends on #228 ?]

xbauch:
This is a refactor of #228. Copy of its description:

This is a partial fix preventing gnat2goto aborting from an unsatisfied pre-condition when a qualified expression is not a numeric subtype.
The current changes work for record subtype qualifiers and numeric subtype qualifiers excluding Integer (and possibly Float - I have not checked this). It does not work for qualified enumeration subtypes. For type Integer and enumeration subtypes it still fails a pre-condition check.
For qualified array subtypes it does not currently perform a check but should not cause a pre-condition failure. Eventually a check of the array subtype must be implemented. At the moment I could not produce a simple test case which got as far as checking an array subtype qualified expression.

The important feature of the partial fix is that it facilitates running list_unsupported script on UKNI without gnat2goto aborting.

I think its shortcomings should be raised as issues.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -1475,14 +1475,27 @@ package body Tree_Walk is
 
    function Do_Qualified_Expression (N : Node_Id) return Irep is
       Value : constant Irep := Do_Expression (Expression (N));
+      Entity_Type : constant Node_Id := Etype (N);
       Type_Of_Val : constant Irep :=
-         Follow_Symbol_Type (
-          Do_Type_Reference (Etype (N)),
-          Global_Symbol_Table);
+        Follow_Symbol_Type (Do_Type_Reference (Entity_Type),
+                            Global_Symbol_Table);
       Typecast_Expr : constant Irep :=
-        Make_Op_Typecast (Value, Sloc (N), Type_Of_Val);
+        Typecast_If_Necessary (Value, Type_Of_Val);
    begin
-      return Make_Range_Assert_Expr (N, Typecast_Expr, Type_Of_Val);
+      --  Todo qualified array expressions - at the moment they will not
+      --  be checked.
+      --  Only perform a range check if the qualification is for a numeric
+      --  enumeration or array subtype
+      if Is_Numeric_Type (Entity_Type)
+        or else Is_Enumeration_Type (Entity_Type)
+      then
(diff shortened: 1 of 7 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -1475,14 +1475,27 @@ package body Tree_Walk is
...
+                            Global_Symbol_Table);
       Typecast_Expr : constant Irep :=
-        Make_Op_Typecast (Value, Sloc (N), Type_Of_Val);
+        Typecast_If_Necessary (Value, Type_Of_Val);
    begin
-      return Make_Range_Assert_Expr (N, Typecast_Expr, Type_Of_Val);
+      --  Todo qualified array expressions - at the moment they will not
+      --  be checked.
+      --  Only perform a range check if the qualification is for a numeric
+      --  enumeration or array subtype
```
hannes-steffenhagen-diffblue: For arrays I guess we'd have to check if it's bounds are matching those given in the (sub)type, if any?

tjj2017: Yes, in the first instance, but we may have to consider sliding.  I will have to check I think its a rather complicated feature of the language.

chrisr-diffblue:
@xbauch @tjj2017 - Again here, could you clarify which of this PR versus #228 needs to be merged and close one or other PRs if they are redundant.

tjj2017:
Having looked at a recent master, this PR seems to have been over taken.  The master contains the following for Do_Qualified_Expression:

      --  TODO: Range expressions for non-bounded types are outside
      --        the scope of this PR
      if Kind (Type_Of_Val) in I_Bounded_Signedbv_Type | I_Bounded_Floatbv_Type
      then
         return Make_Range_Assert_Expr (N, Typecast_Expr, Type_Of_Val);
      else
         return Typecast_Expr;
      end if;
I think this will either perform a check for a limited sort of types or ignore the check.
This solves the immediate problem of nor handling qualified expressions but I don't know under which PR this change was incorporated.

xbauch:
@chrisr-diffblue : As @tjj2017 writes above, PR #242 included the `if` branch so we handle qualified expression but do not perform range check for non-bounded types. I think this PR can be closed.

chrisr-diffblue:
Thanks for the clarifications. Closing this PR now.

PR #234, opened 2019-05-17
Ignore warnings in regression tests

xbauch:
Includes a regression test that generates a warning.

Reasoning: System package contains export/import pragmas that we want to ignore
but now about. Our regression test was sensitive to all gnat2goto output,
including warnings.


tjj2017:
The warnings are generated by the Import pragmas.  A warning is given by gnat2goto that non Ada languages are not supported.  However, in the case of the Import pragmas in System and its child packages, the calling convention is "intrinsic" which indicates it is a subprogram built into the compiler (see ALRM 6.3.1) and as such I think can be safely ignored without issuing a warning.

I have a draft modification which does this but have not committed it yet.  I wanted to get other people's views.

xbauch:
Unnecessary now that #237 is merged (and does not produce any warnings).

PR #235, opened 2019-05-17
Feature cbmc bump

hannes-steffenhagen-diffblue:
Bump CBMC and fix issues arising from it

### DIFF
--- a/gnat2goto/driver/driver.adb
+++ b/gnat2goto/driver/driver.adb
@@ -145,6 +145,33 @@ package body Driver is
    procedure Initialize_CProver_Internal_Variables (Start_Body : Irep) is
       Int_32_T : constant Irep := Make_Signedint_Type (32);
 
+      procedure Declare_Missing_Global (Symbol_Expr : Irep)
+        with Pre => Kind (Symbol_Expr) = I_Symbol_Expr;
+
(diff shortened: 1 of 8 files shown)


gnat2goto/driver/driver.adb
```diff
@@ -164,6 +191,75 @@ package body Driver is
          Append_Op (Start_Body, Initialization_Statement);
       end Initialize_CProver_Rounding_Mode;
 
+      procedure Initialize_CProver_Dead_Object;
```
xbauch: So these three procedure only differ in the `Identifier`?

hannes-steffenhagen-diffblue: pretty much. I can abstract if if you prefer?

xbauch: I'm not insisting.

chrisr-diffblue:
I've confirmed with @martin-cs that it's ok to bump the submodule - so just needs CI sorting and second review.

chrisr-diffblue:
BTW - If there is a specific reason for the submodule bump, (e.g. to pickup a fix/feature we need) it would be fantastic if that reason was included in the commit message for the commit that actually moves the submodule pointer :-)

PR #236, opened 2019-05-17
Ignore preelaborate initialization and warnings

xbauch:
Two more pragmas to ignore. These two need to be handled if we want to process `System` package.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4856,9 +4856,6 @@ package body Tree_Walk is
          when Name_Unreferenced =>
             Report_Unhandled_Node_Empty (N, "Process_Pragma_Declaration",
                                          "Unsupported pragma: Unreferenced");
-         when Name_Preelaborable_Initialization =>
-            Report_Unhandled_Node_Empty (N, "Process_Pragma_Declaration",
-                           "Unsupported pragma: Preelaborable Initialization");
          when Name_Ada_05 =>
             Report_Unhandled_Node_Empty (N, "Process_Pragma_Declaration",
                                          "Unsupported pragma: Ada 05");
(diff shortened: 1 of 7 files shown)


xbauch:
@martin-cs What do you think?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4946,8 +4940,12 @@ package body Tree_Walk is
             --  Ignored for now
               Name_No_Elaboration_Code_All |
             --  Only affects elaboration and linking so can be ignored for now
-              Name_Universal_Aliasing =>
+              Name_Universal_Aliasing |
             --  Optimisation control, should be ignored
+              Name_Preelaborable_Initialization |
```
hannes-steffenhagen-diffblue: FWIW I don't think ignoring these is right. We _will_ need to handle these eventually.

xbauch: Well they are part of UKNI, so eventually might as well be now.

tjj2017: Preelaborate is used to specify and control elaboration before calling the subprogram designated as the main program.  The conditions that allow a pragma preelaborate to be validly applied to a unit is checked by the front end.  I am hoping to use the gnat binder to define the elaboration order and again that would require no action from gnat2goto.  I think pragma Preelaborate could safely be ignored by gnat2goto.  gnat2goto may have to produce initialisation routines which facilitate elaboration but not the order in which they are called.

tjj2017:
Preelaborate Initialisation causes the front end to do checks to ensure the application of the pragma is valid.  It may affect the elaboration order but that should not implicate gnat2goto.  I think this pragma can be safely ignored too.

martin-cs:
@xbauch sorry for not responding sooner.

@hannes-steffenhagen-diffblue is right that we will need to handle them.

@tjj2017 is right that it is not a front-end issue and should be handled when linking, so it is safe to ignore them in the front-end

chrisr-diffblue:
I'll rebase this in a new PR.

chrisr-diffblue:
Merged in #249 so closing this PR now.

PR #231, opened 2019-05-17
Define package body symbol

xbauch:
This is a refactor of #230 with updated golden results. Copy of its description:

In Do_Compilation_Unit the return value Unit_Symbol was not set for a package body

To determine a symbol value for a package body a symbol is determined from
its specification and registered in the symbol table.  This required checking for
the package being a child.

Note: The same check may be required for child subprograms!

PR #237, opened 2019-05-17
Process system package

xbauch:
Start processing the `System` package.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4369,7 +4369,8 @@ package body Tree_Walk is
    procedure Do_Withed_Unit_Spec (N : Node_Id) is
    begin
       if Defining_Entity (N) = Stand.Standard_Standard then
-         --  At the moment Standard is not processed: TODO
+         --  TODO: github issue #252
+         --  At the moment Standard is not processed
          null;
       else
          --  Handle all other withed library unit declarations
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -1,4 +1,3 @@
-with Uname;                 use Uname;
 with Namet;                 use Namet;
 with Nlists;                use Nlists;
 with Sem;
@@ -1028,6 +1027,9 @@ package body Tree_Walk is
                Dummy : constant Irep := Do_Subprogram_Or_Block (U);
                pragma Unreferenced (Dummy);
             begin
+            --  The specification of the package body has already
+            --  been inserted into the symbol table by the call to
+            --  Do_Withed_Unit_Specs.
                pragma Assert (Global_Symbol_Table.Contains (Unit_Name));
                Unit_Symbol := Global_Symbol_Table (Unit_Name);
                Unit_Is_Subprogram := False;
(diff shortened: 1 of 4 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -3501,7 +3502,35 @@ package body Tree_Walk is
...
+
+      Package_Name := Intern (Unique_Name (Entity_Node));
+      Package_Symbol.Name       := Package_Name;
+      Package_Symbol.BaseName   := Package_Name;
+      Package_Symbol.PrettyName := Package_Name;
+      Package_Symbol.SymType    := New_Irep (I_Void_Type);
+      Package_Symbol.Mode       := Intern ("C");
+      Package_Symbol.Value      := Make_Nil (Sloc (N));
+
+      Global_Symbol_Table.Insert (Package_Name, Package_Symbol);
```
hannes-steffenhagen-diffblue: Why are we inserting the package into the symbol table? :question:

testsuite/gnat2goto/tests/abs/test.out
```diff
@@ -1,7 +1,19 @@
+Standard_Error from gnat2goto absolute:
+Warning: Multi-language analysis unsupported.
```
hannes-steffenhagen-diffblue: for this it might be worth to add something like

```
prove(ignore_lines=[
  'Warning: Multi-language analysis unsupported.'
  ])
```

which would drop all lines matching some pattern from the output

gnat2goto/driver/tree_walk.adb
xbauch: I guess this refers to #231.

tjj2017: To avoid an initialised symbol in Do_Compilation_Unit

testsuite/gnat2goto/tests/abs/test.out
tjj2017: See my comment about ignoring pragma Import when the calling convention is "Intrinsic" in PR#234

tjj2017: Another option is to use the standard warning mechanism of the gnat front end to generate warning messages and then we could use the pragma Warnings to control reporting.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4365,12 +4367,9 @@ package body Tree_Walk is
...
 
    procedure Do_Withed_Unit_Spec (N : Node_Id) is
-      Unit_Name : constant String := Get_Name_String (Get_Unit_Name (N));
    begin
-      if Defining_Entity (N) = Stand.Standard_Standard or else
-        Unit_Name = "system%s"
-      then
-         --  At the moment Standard or System are not processed: TODO
+      if Defining_Entity (N) = Stand.Standard_Standard then
+         --  At the moment Standard is not processed: TODO
```
chrisr-diffblue: Could you raise a GitHub issue to cover this, and then reference the issue in the TODO comment here?

xbauch: #252 and also added to the comment.

xbauch:
@NlightNFotis before in `do_withed_spec` we ignored two kinds of packages
1: those with defining identity being `Standard_Standard`
2: those whose name had prefix `system`
this commit stops ignoring the system packages and runs the same processing on them (of the withed spec) as for user-defined packages.

Issue #226, opened 2019-05-15
Range checks for subtypes not supported correctly.


NlightNFotis:
This has been a problem that I discovered while working on supporting subtypes within enumeration symbols.

For example, this code is failing:

```ada
procedure Range_Second is
  type Indicator_Type is (
    CALIBRATION_MODE,
    MEASUREMENT_MODE,
    CALIBRATION_PASS,
    CALIBRATION_FAIL,
    MEASUREMENT_NOT_PROVEN,
    MEASUREMENT_PRESENT);

  subtype Result_Indicator_Type is Indicator_Type range CALIBRATION_PASS .. MEASUREMENT_NOT_PROVEN;

  procedure Testable (Result : Result_Indicator_Type) is
  begin
    null;
  end Testable;

begin
  Testable (CALIBRATION_PASS);
  Testable (MEASUREMENT_PRESENT);
end Range_Second;
```

But the problem is not necessarily on the symbols signifying the range, as the same program with integer ranges is also failing the range checks.

```ada
procedure Range_Second is
  type An_Int is range 1..1000;
  subtype Another_Int is An_Int range 5..100;

  procedure Testable (Result : Another_Int) is
  begin
    null;
  end Testable;

begin
  Testable (50);
  Testable (200);
end Range_Second;
```

Failing the range checks in this case means that CBMC returns `VERIFICATION SUCCESSFUL` when otherwise it should have assertion failures.

PR #238, opened 2019-05-17
Fix UKNI range issue, and add symbols earlier to the symbol table

NlightNFotis:
Fix the range exception problem for UKNI, and also fix another bug where symbols where not being added to the symbol table at the appropriate stage.

This contains two tests that are not fully functional right now due to the problem described in #226


gnat2goto/driver/tree_walk.adb
```diff
@@ -792,23 +805,44 @@ package body Tree_Walk is
...
+               Val : constant Irep := Global_Symbol_Table
+                 (Intern
+                    (Get_Identifier
+                     (Mem)))
+                 .Value;
+            begin
+               return Val;
+            end;
+         else
+            return Mem;
```
xbauch: So the function is identity for non-`C_Enum_Type`?

gnat2goto/driver/tree_walk.adb
```diff
@@ -792,23 +805,44 @@ package body Tree_Walk is
...
+         Followed_Type_Symbol : constant Irep :=
+            Follow_Symbol_Type (Get_Type (Mem), Global_Symbol_Table);
+      begin
+         if Kind (Followed_Type_Symbol) = I_C_Enum_Type then
+            declare
+               Val : constant Irep := Global_Symbol_Table
+                 (Intern
+                    (Get_Identifier
+                     (Mem)))
+                 .Value;
```
xbauch: So this function returns the initial value for the symbol? What if it's modified?

gnat2goto/driver/tree_walk.adb
NlightNFotis: Yes. Do you think it's better to be documented? I thought it was obvious.

gnat2goto/driver/tree_walk.adb
NlightNFotis: I will need to get back to you on that.

gnat2goto/driver/tree_walk.adb
xbauch: No. I was just trying to find out if it was intentional.

NlightNFotis:
@xbauch I updated the golden results, but something in them doesn't look right. Can you also take a look if there's something out of place?

gnat2goto/driver/tree_walk.adb
```diff
@@ -792,23 +805,44 @@ package body Tree_Walk is
 
       procedure Handle_Parameter (Formal : Entity_Id; Actual : Node_Id);
 
+      function Handle_Enum_Symbol_Members (Mem : Irep) return Irep;
```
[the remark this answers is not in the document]
NlightNFotis: This refers to the `Member`.

testsuite/gnat2goto/tests/range_second/test.opt
```diff
@@ -0,0 +1,2 @@
+ALL XFAIL Doesn't crash anymore but cbmc does not perform range assertion checks
```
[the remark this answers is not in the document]
NlightNFotis: I don't think there's any relevant test for this. The range of this PR was for a very particular piece of code (using enum symbols as range constraints) to not crash. This has been fixed, but gnat2goto suffers from a different problem now (it doesn't handle constraints of subtypes correctly). I can't easily think of something that can exercise the fix without falling on its face because of the subtype issue.

gnat2goto/driver/tree_walk.adb
```diff
@@ -2882,6 +2916,24 @@ package body Tree_Walk is
 
       --  Begin processing for Do_Object_Declaration_Full_Declaration
 
+      procedure Register_Identifier_In_Symbol_Table (N : Irep; Val : Irep)
```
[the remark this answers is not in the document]
NlightNFotis: There's nothing there now, but I remember a PR by Petr that was adding like a couple more similar functions. They should be refactored in the future into one, but not as part of this PR.

gnat2goto/driver/tree_walk.adb
```diff
@@ -765,13 +770,21 @@ package body Tree_Walk is
...
+            Upper_Bound_Value :=
+               Store_Symbol_Bound (Bound_Type_Symbol (Upper_Bound));
          when others =>
             Report_Unhandled_Node_Empty (Upper_Bound,
                                          "Do_Base_Range_Constraint",
                                          "unsupported upper range kind");
       end case;
 
-      Set_Width (Result_Type, Get_Width (Resolved_Underlying));
+      if Kind (Resolved_Underlying) = I_C_Enum_Type then
```
hannes-steffenhagen-diffblue: This seems sketchy. `enumeration_typet` does not have a width field, seems like you're basically invoking UB here?

gnat2goto/driver/tree_walk.adb
```diff
@@ -2882,6 +2918,24 @@ package body Tree_Walk is
...
+            Intern (Get_Identifier (N));
+         Identifier_Symbol : Symbol;
+      begin
+         Identifier_Symbol.Name       := Identifier_Name;
+         Identifier_Symbol.BaseName   := Identifier_Name;
+         Identifier_Symbol.PrettyName := Identifier_Name;
+         Identifier_Symbol.SymType    := Get_Type (N);
+         Identifier_Symbol.Mode       := Intern ("C");
+         Identifier_Symbol.Value      := Val;
+         Global_Symbol_Table.Insert (Identifier_Name, Identifier_Symbol);
```
hannes-steffenhagen-diffblue: Should have `IsLValue => True` and `IsStaticLifetime => True`

gnat2goto/driver/tree_walk.adb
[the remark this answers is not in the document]
hannes-steffenhagen-diffblue: This should probably go into the symbol table. I'm not sure about the arguments either; Rather than having an `Irep` and requiring ito to be a `Symbol_Expr`, why not pass in the `name` and `type` (which are ultimately what you need) as parameters?

Should also be called something like `Declare_Global_Variable`?

Also this should probably go into `goto_utils`

gnat2goto/driver/tree_walk.adb
```diff
@@ -2907,6 +2961,11 @@ package body Tree_Walk is
                                              Rhs => Init_Expr,
                                              Source_Location => Sloc (N)));
       end if;
+
+      if not Is_In_Symtab then
+         Register_Identifier_In_Symbol_Table (Id, Init_Expr);
```
hannes-steffenhagen-diffblue: I'm not sure I quite understand what what this here does?

NlightNFotis: What this does is that there was a problem where symbols where not added to the symbol table with the values they should have at one path that handled object declarations in the past, and these caused symbol table lookups to fail. This was fixing this.

gnat2goto/driver/tree_walk.adb
NlightNFotis: The `enumeration_typet` doesn't have a `width` field, but it has a subtype that has a `width` field.

gnat2goto/driver/tree_walk.adb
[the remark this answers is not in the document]
NlightNFotis: Again, because a number of similar functions were introduced recently for very similar reasons, just in different paths, I suggest that this be done as a more organised attempt to refactor these functions and this one, and make sure that the refactored version gets appropriately invoked in the correct places. But I don't think it's within the scope of this PR for this to happen.

PR #239, opened 2019-05-20
Representation clause component size

xbauch:
In this PR we only check that the component size is the same as the width of the related subtype. That seems to be enough for UKNI. There is also a failing regression test that tries the change the component size.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4589,21 +4589,22 @@ package body Tree_Walk is
       procedure Handle_Representation_Clause (N : Node_Id) is
          Attr_Id : constant String := Get_Name_String (Chars (N));
          Target_Name : constant Irep := Do_Identifier (Name (N));
-         Entity_Esize : constant Integer :=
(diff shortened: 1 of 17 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4554,9 +4554,31 @@ package body Tree_Walk is
...
+                 Follow_Symbol_Type (Get_Subtype (Get_Type (Array_Data)),
+                                     Global_Symbol_Table);
+               Target_Subtype_Width : constant Uint :=
+                 UI_From_Int (Int (Get_Width (Target_Subtype)));
+            begin
+               if Component_Size (Entity (N)) /= Expression_Value or
+                 Target_Subtype_Width /= Expression_Value
+               then
+                  Report_Unhandled_Node_Empty (N, "Process_Declaration",
+                              "Changing component size currently unsupported");
```
hannes-steffenhagen-diffblue: Maybe `"Having component sizes be different from the size of their underlying type is currently not supported"`?

PR #240, opened 2019-05-21
Refactor symtab inserts

xbauch:
We were adding into the symbol table in a number of places across `tree_walk`. This PR wraps every access into a helper function in `goto_utils`. Checks that declared objects are present in the symbol table and introduces them otherwise. In the last bit, there is an overlap with #238.

### DIFF
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -121,6 +121,23 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Enum_Member_Symbol_Entry (
+      Member_Name : Symbol_Id; Base_Name : Symbol_Id; Enum_Type : Irep;
+      Value_Expr : Irep; A_Symbol_Table : in out Symbol_Table) is
+      Member_Symbol : Symbol;
+   begin
+      Member_Symbol.Name             := Member_Name;
+      Member_Symbol.PrettyName       := Base_Name;
+      Member_Symbol.BaseName         := Base_Name;
+      Member_Symbol.Mode             := Intern ("C");
+      Member_Symbol.IsStaticLifetime := True;
+      Member_Symbol.IsStateVar       := True;
+      Member_Symbol.SymType          := Enum_Type;
+      Member_Symbol.Value            := Value_Expr;
+
+      A_Symbol_Table.Insert (Member_Symbol.Name, Member_Symbol);
+   end New_Enum_Member_Symbol_Entry;
+
    --------------------------------
    -- New_Parameter_Symbol_Entry --
    --------------------------------
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -121,6 +121,20 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Type_Symbol_Entry (Type_Name : Symbol_Id; Type_Of_Type : Irep;
+                                    A_Symbol_Table : in out Symbol_Table) is
+      Type_Symbol : Symbol;
+   begin
+      Type_Symbol.SymType    := Type_Of_Type;
+      Type_Symbol.IsType     := True;
+      Type_Symbol.Name       := Type_Name;
+      Type_Symbol.PrettyName := Type_Name;
+      Type_Symbol.BaseName   := Type_Name;
+      Type_Symbol.Mode       := Intern ("C");
+
+      A_Symbol_Table.Insert (Type_Name, Type_Symbol);
+   end New_Type_Symbol_Entry;
+
    procedure New_Valueless_Object_Symbol_Entry (Constant_Name : Symbol_Id;
                                         A_Symbol_Table : in out Symbol_Table)
    is
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -121,6 +121,21 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Valueless_Object_Symbol_Entry (Constant_Name : Symbol_Id;
+                                        A_Symbol_Table : in out Symbol_Table)
+   is
+      Object_Symbol : Symbol;
+   begin
+      Object_Symbol.Name       := Constant_Name;
+      Object_Symbol.BaseName   := Constant_Name;
+      Object_Symbol.PrettyName := Constant_Name;
+      Object_Symbol.SymType    := Make_Nil (No_Location);
+      Object_Symbol.Mode       := Intern ("C");
+      Object_Symbol.Value      := Make_Nil (No_Location);
+
+      A_Symbol_Table.Insert (Constant_Name, Object_Symbol);
+   end New_Valueless_Object_Symbol_Entry;
+
    procedure New_Enum_Member_Symbol_Entry (
       Member_Name : Symbol_Id; Base_Name : Symbol_Id; Enum_Type : Irep;
       Value_Expr : Irep; A_Symbol_Table : in out Symbol_Table) is
--- a/gnat2goto/driver/goto_utils.adb
+++ b/gnat2goto/driver/goto_utils.adb
@@ -121,6 +121,24 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Object_Symbol_Entry (Object_Name : Symbol_Id;
+                                      Object_Type : Irep;
+                                      Object_Init_Value : Irep;
+                                      A_Symbol_Table : in out Symbol_Table)
+   is
+      Object_Symbol : Symbol;
+   begin
+      Object_Symbol.Name       := Object_Name;
+      Object_Symbol.BaseName   := Object_Name;
+      Object_Symbol.PrettyName := Object_Name;
+      Object_Symbol.SymType    := Object_Type;
+      Object_Symbol.Mode       := Intern ("C");
+      Object_Symbol.Value      := Object_Init_Value;
+      Object_Symbol.IsLValue   := True;
+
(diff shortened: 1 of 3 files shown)


gnat2goto/driver/goto_utils.adb
```diff
@@ -120,6 +121,22 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Constant_Symbol_Entry (N : Node_Id;
```
hannes-steffenhagen-diffblue: If the `Node_Id` is just used to get a name and a source location, can we just pass those to the function instead? The function doesn't seem node specific.

gnat2goto/driver/goto_utils.ads
```diff
@@ -40,6 +40,12 @@ package GOTO_Utils is
    function Symbol_Expr (Sym : Symbol) return Irep
      with Post => Kind (Symbol_Expr'Result) = I_Symbol_Expr;
 
+   procedure New_Constant_Symbol_Entry (N : Node_Id;
+                                        A_Symbol_Table : in out Symbol_Table)
+     with Pre => Nkind (N) = N_Object_Declaration;
+     --  Adds a dummy entry to the symbol table to register that a
+     --  constant has already been processed.
```
hannes-steffenhagen-diffblue: What does this mean?

gnat2goto/driver/goto_utils.adb
```diff
@@ -121,6 +121,19 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Type_Symbol_Entry (Type_Name : String; Type_Type : Irep;
```
hannes-steffenhagen-diffblue: `Type_Type` type type type type.

gnat2goto/driver/goto_utils.ads
```diff
@@ -40,6 +40,10 @@ package GOTO_Utils is
    function Symbol_Expr (Sym : Symbol) return Irep
      with Post => Kind (Symbol_Expr'Result) = I_Symbol_Expr;
 
+   procedure New_Type_Symbol_Entry (Type_Name : String; Type_Type : Irep;
```
hannes-steffenhagen-diffblue: The `Type_Name` is only used `Intern`'ed, so IMHO this should probably just use a `Symbol_Id`

gnat2goto/driver/tree_walk.adb
```diff
@@ -4427,15 +4428,14 @@ package body Tree_Walk is
             Set_Return_Type (Fn_Type, Void_Type);
             Set_Assertion (Assertion, Formal_Expr);
 
-            Fn_Symbol.Name := Intern ("__ada_runtime_check");
-            Fn_Symbol.PrettyName := Fn_Symbol.Name;
-            Fn_Symbol.BaseName := Fn_Symbol.Name;
-            Fn_Symbol.Value := Assertion;
-            Fn_Symbol.SymType := Fn_Type;
-            Global_Symbol_Table.Insert (Fn_Symbol.Name, Fn_Symbol);
+            Fn_Symbol :=
```
hannes-steffenhagen-diffblue: Can this be moved up to the declaration, and the symbol be made constant now?

gnat2goto/driver/goto_utils.adb
```diff
@@ -121,6 +121,23 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Object_Symbol_Entry (Object_Name : Symbol_Id;
```
hannes-steffenhagen-diffblue: What's "object" in this context? We don't usually use that terminology in CBMC. I believe that's what GNAT calls an LValue? If so, this should have IsLValue set to true.

gnat2goto/driver/goto_utils.adb
```diff
@@ -120,6 +120,23 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Member_Symbol_Entry (Member_Name : String; Base_Name : String;
+                                      Enum_Type : Irep; Value_Expr : Irep;
+                                      A_Symbol_Table : in out Symbol_Table) is
+      Member_Symbol : Symbol;
+   begin
+      Member_Symbol.Name := Intern (Member_Name);
+      Member_Symbol.PrettyName := Intern (Base_Name);
```
NlightNFotis: Why are the symbols getting interned here? In all the similar implementations of this we had the symbols where not interned.

gnat2goto/driver/goto_utils.adb
```diff
@@ -120,6 +120,23 @@ package body GOTO_Utils is
       return Ret;
    end Symbol_Expr;
 
+   procedure New_Member_Symbol_Entry (Member_Name : String; Base_Name : String;
```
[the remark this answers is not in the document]
NlightNFotis: Element of a product type?

gnat2goto/driver/goto_utils.ads
NlightNFotis: Petr didn't add this, this was already there, and I agree that it was weird.

hannes-steffenhagen-diffblue: I can see that, but if we're changing this already my opinion is

A) Either we figure out what this means and reword it so it's easier to understand
B) If we can't figure out what this means between the 3 of us I'd posit it's clearly not a very useful comment, so I'd rather remove it in that case.

gnat2goto/driver/goto_utils.adb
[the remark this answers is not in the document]
xbauch: I was trying to do this PR with unnecessary renaming. Previously, we created a `member_symbol` and inserted it into the symbol table.

gnat2goto/driver/goto_utils.adb
xbauch: Copy pasted. Also wouldn't type-check, `name` is `symbol_id` not a string.

gnat2goto/driver/goto_utils.ads
```diff
@@ -38,7 +38,12 @@ package GOTO_Utils is
         Post => Kind (Param_Symbol'Result) = I_Symbol_Expr;
 
    function Symbol_Expr (Sym : Symbol) return Irep
-   with Post => Kind (Symbol_Expr'Result) = I_Symbol_Expr;
+     with Post => Kind (Symbol_Expr'Result) = I_Symbol_Expr;
+
+   procedure New_Member_Symbol_Entry (Member_Name : String; Base_Name : String;
+                                     Enum_Type : Irep; Value_Expr : Irep;
```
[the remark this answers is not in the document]
xbauch: Well before it was `enum_type_symbol`. I can re-name it back if you want.

gnat2goto/driver/goto_utils.ads
```diff
@@ -38,7 +38,12 @@ package GOTO_Utils is
...
+                                     A_Symbol_Table : in out Symbol_Table)
+     with Pre => Kind (Enum_Type) = I_Symbol_Type;
```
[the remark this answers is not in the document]
xbauch: It probably doesn't. I thought: if somebody finds a different use-case for this procedure, they'd have to explicitly change it.

gnat2goto/driver/goto_utils.adb
xbauch: Yes, `node_id` is used for the same purpose as in the original procedure.

gnat2goto/driver/goto_utils.adb
xbauch: Well before it was `actual_type`. I can re-name it back.

gnat2goto/driver/tree_walk.adb
xbauch: Well we are changing the `assertion` that will comprise the `value` of this symbol.

gnat2goto/driver/goto_utils.adb
xbauch: The `object` from `do_object_declaration`.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1221,7 +1221,7 @@ package body Tree_Walk is
               UI_Image (Enumeration_Rep (Member));
             Val_Name : constant String := Unique_Name (Member);
             Base_Name : constant String := Get_Name_String (Chars (Member));
-            Member_Symbol : Symbol;
+
```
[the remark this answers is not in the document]
xbauch: Un-removed.

gnat2goto/driver/goto_utils.ads
xbauch: Maybe the comment makes more sense in the call-site: we call it when declaring a constant object, that does not have initial expression -- deferred constant declaration. It's completion is somewhere down the tree but we do not want to process is again which is why store it in the symbol table without the value it will be eventually initialised with.

I moved the comment to the call-site.

gnat2goto/driver/goto_utils.ads
xbauch: Intern was inlined at call-site.

gnat2goto/driver/goto_utils.ads
xbauch: I also renamed the procedure to `New_Valueless_Object_Symbol_Entry`.

gnat2goto/driver/goto_utils.adb
```diff
@@ -154,52 +154,51 @@ package body GOTO_Utils is
       A_Symbol_Table.Insert (Subprog_Name, Subprog_Symbol);
    end New_Subprogram_Symbol_Entry;
 
-   procedure New_Type_Symbol_Entry (Type_Name : String; Type_Type : Irep;
+   procedure New_Type_Symbol_Entry (Type_Name : Symbol_Id; Type_Of_Type : Irep;
```
hannes-steffenhagen-diffblue: That kind of makes it sound like we have higher kinded types? I honestly don't really have a fantastic idea of what to name this either though.

PR #241, opened 2019-05-21
Do not analyse children of system

tjj2017:
Rather than re-instate the analysis of package system, this change inhibits the analysis of packages standard and system and all their children.

I don't this is a final solution and I would like to put out a message of unsupported node but every unit automatically with's Standard and some units implicitly with System even though it is not explicitly used.

Open for discussion.


tjj2017:
I am closing this pull request because gnat2goto is well on the way to being able to handle package system and its child packages.

PR #243, opened 2019-05-30
Handle intrinsic import

tjj2017:
A pragma Import with the convention Intrinsic does not signify a multi language program.  Intrinsic convention is used to indicate a subprogram built in to the Ada compiler.  It can be safely ignored by gnat2goto but it will not necessarily have a body in the translated code.


testsuite/gnat2goto/tests/intrinsic/use_import.adb
```diff
@@ -0,0 +1,17 @@
...
+
+   function "-" (X : Integer) return Integer;
+   pragma Import (Convention => Intrinsic, Entity => "-");
+
+   function "+" (Left, Right : Integer) return Integer;
+   pragma Import (Intrinsic, "+");
+
+   I : Integer := 1;
+begin
+   I := -I;
```
hannes-steffenhagen-diffblue: What is this for?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4814,8 +4814,40 @@ package body Tree_Walk is
             --  be called from Ada, or a foreign-language variable to be
             --  accessed from Ada. This would (probably) require gnat2goto to
             --  understand the foreign code, which we do not at the moment.
```
hannes-steffenhagen-diffblue: FWIW this bit isn't actually true anymore; Since we now support linking of symtab files we _can_ actually do a multi language analysis just fine, we just have to make sure the name of the functions match.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4814,8 +4814,40 @@ package body Tree_Walk is
...
-            Put_Line (Standard_Error,
-                      "Warning: Multi-language analysis unsupported.");
+            --  However, if the calling convention is specified as "Intrinsic"
+            --  then the subprogram is built into the compiler and gnat2goto
+            --  can safely ignore the pragma.
+            declare
+               --  If the pragma is specified with positional parameter
+               --  association, then the calling convetion is the first
```
NlightNFotis: small typo here "convention"

martin-cs:
On Fri, 2019-05-31 at 09:11 -0700, hannes-steffenhagen-diffblue wrote:
> @@ -4814,8 +4814,40 @@ package body Tree_Walk is
> 
>              --  be called from Ada, or a foreign-language variable
> to be
>              --  accessed from Ada. This would (probably) require
> gnat2goto to
>              --  understand the foreign code, which we do not at the
> moment.
> 
> FWIW this bit isn't actually true anymore; Since we now support
> linking of symtab files we _can_ actually do a multi language
> analysis just fine, we just have to make sure the name of the
> functions match.

Hannes is right; it would not necessarily require gnat2goto
understanding the other language.  However I also wouldn't say we can
yet do multi-language analysis as this will also require some work on
calling conventions and how variables and types are mapped between the
two.

We have a *route* to supporting multi-language analysis via separate
compilation and linking but we have not yet developed it.

tjj2017:
I have rebased this to diffblue/master

testsuite/gnat2goto/tests/intrinsic/use_import.adb
tjj2017: It calls the intrinsic function "-"

gnat2goto/driver/tree_walk.adb
tjj2017: Corrected

chrisr-diffblue:
Supreceeded by #251

PR #245, opened 2019-06-01
Subprogram body stub

tjj2017:
UKNI has subprogram_body_stubs.  The changes on this branch facilitate handling of these stubs by gnat2goto.

The test program which has been added to the regression tests for this new feature, P, has an assertion I think should succeed (if overflows are ignored), but it seems at the moment that the assertion fails.  The action of the called subprogram, Inc, does not seem to be considered by cbmc.

Should I mark this with an XFAIL?  If so I think I should have another test without the assertion so that the subprogram_body_stub feature is always tested.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -4143,9 +4143,9 @@ package body Tree_Walk is
 
    procedure Do_Subprogram_Body_Stub (N : Node_Id) is
    begin
-      --  The Gnat compilation model requires that a file containing the
-      --  which is the separate subprogram body is present otherwise
-      --  a compilation error is generated.
+      --  The Gnat compilation model requires that a file
+      --  containing the separate subprogram body is present
+      --  otherwise a compilation error is generated.
       --  Therefore, the subunit will always be present when gnat2goto
       --  encounters a Subprogram_Body_Stub.
       Do_Subprogram_Body (Proper_Body (Unit ((Library_Unit (N)))));
@@ -4779,8 +4779,6 @@ package body Tree_Walk is
             --  body_stub  --
 
          when N_Subprogram_Body_Stub =>
---            Report_Unhandled_Node_Empty (N, "Process_Declaration",
---                                         "Subprogram body stub declaration");
             Do_Subprogram_Body_Stub (N);
 
          when N_Package_Body_Stub =>
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -252,6 +252,9 @@ package body Tree_Walk is
    procedure Do_Subprogram_Body (N : Node_Id)
    with Pre => Nkind (N) = N_Subprogram_Body;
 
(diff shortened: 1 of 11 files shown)


tjj2017:
I have rebased this branch to diffblue/master

gnat2goto/driver/tree_walk.adb
```diff
@@ -4723,8 +4740,9 @@ package body Tree_Walk is
             --  body_stub  --
 
          when N_Subprogram_Body_Stub =>
-            Report_Unhandled_Node_Empty (N, "Process_Declaration",
-                                         "Subprogram body stub declaration");
+--            Report_Unhandled_Node_Empty (N, "Process_Declaration",
+--                                         "Subprogram body stub declaration");
```
xbauch: Please remove rather then comment out.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4094,7 +4097,21 @@ package body Tree_Walk is
...
-   -------------------------------
+   -----------------------------
+   -- Do_Subprogram_Body_Stub --
+   -----------------------------
+
+   procedure Do_Subprogram_Body_Stub (N : Node_Id) is
+   begin
+      --  The Gnat compilation model requires that a file containing the
+      --  which is the separate subprogram body is present otherwise
+      --  a compilation error is generated.
```
xbauch: I'm not sure I understand this sentence.

testsuite/gnat2goto/tests/separate_subprog/test.out
```diff
@@ -0,0 +1,3 @@
+[2] file p-inc.asu line 6 assertion: SUCCESS
+[1] file p.adb line 14 assertion: FAILURE
```
xbauch: Why do we want this to fail?

gnat2goto/driver/tree_walk.adb
tjj2017: Removed comments

gnat2goto/driver/tree_walk.adb
tjj2017: Corrected comment

testsuite/gnat2goto/tests/separate_subprog/test.out
tjj2017: We do not want it to fail.  It should succeed if the possibility of an overflow is ignored.  I have split the test into 2.  One with tis result so that we can test for regressions and one with the correct result with I have marked with an XFAIL.

tjj2017:
I have rebased this on PR #220 as this fix is required for this branch.  It passes the regression tests and I have updated the golden results

martin-cs:
Should that be rebased on to #256 ?

chrisr-diffblue:
@martin-cs - yeah, it's been rebased on master now #256 has been merged.

testsuite/gnat2goto/tests/separate_subprog/test.out
xbauch: Thanks.

PR #248, opened 2019-06-03
Feature/store defined symbols

chrisr-diffblue:


chrisr-diffblue:
Now redundant because the original PR, #220 has been merged.

PR #250, opened 2019-06-04
In Do_Compilation_Unit the return value Unit_Symbol was not set for a…

tjj2017:
… package body

To determine a symbol value for a package body a symbol is determined from
its specification and registered in the symbol table.  This required checking for
the package being a child.
Note: The same check may be required for child subprograms!


tjj2017:
As suggested by Chris, here is a new pull request for this branch.

chrisr-diffblue:
The commits in this PR were incorporated into #231, which has now been merged, therefore this PR has been superseded.

PR #255, opened 2019-06-13
List unsupported user help

chrisr-diffblue:
This PR is probably easier to review commit-by-commit, rather than as a whole diff. There is only a small amount of real change, but some code has moved around making the whole diff for the PR look more messy than it really is.

### DIFF
--- a/experiments/list_unsupported.sh
+++ b/experiments/list_unsupported.sh
@@ -27,21 +27,6 @@ if [ "$1" = '--help' ]; then
    exit 3
 fi
 
-echo >&2 "Project to build: $1"
-path="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
-include_path=""
-file_name=$(basename "$1")
-
-for foldername in $(find ${path} -type d -name "*" | LC_ALL=posix sort ); do
-   count=`ls -1 ${foldername}/*.ads 2>/dev/null | wc -l`
-   if [ $count != 0 ]
-   then
-      include_path="${include_path} -I ${foldername}"
-   fi
-done
-
-echo "$1: Unsupported features\n" > "$file_name".txt
-
 # gnat on the path?
 if ! command -v gnat > /dev/null; then
    echo >&2 "Gnat not on PATH!"
@@ -110,6 +95,21 @@ export PATH="${saved_path}"
 echo >&2 "...environment is OK."
 
 # Finally start work
+
+echo >&2 "Project to build: $1"
+path="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
+include_path=""
+file_name=$(basename "$1")
+
+for foldername in $(find ${path} -type d -name "*" | LC_ALL=posix sort ); do
+   count=`ls -1 ${foldername}/*.ads 2>/dev/null | wc -l`
+   if [ $count != 0 ]
+   then
+      include_path="${include_path} -I ${foldername}"
+   fi
+done
+
+echo "$1: Unsupported features\n" > "$file_name".txt
 
 # Enumerate all the sub directories of ADA_INCLUDE_PATH
 for include_folder in `echo "$ADA_INCLUDE_PATH" | tr ':' ' '` ; do
--- a/experiments/list_unsupported.sh
+++ b/experiments/list_unsupported.sh
@@ -61,19 +61,33 @@ if ! (gnat --version | grep -q 'GNAT GPL 2016') ; then
    exit 5
 fi
 
+# Sane ADA_INCLUDE_PATH and GPR_PROJECT_PATH?
 GNAT2GOTO=`command -v gnat2goto`
 ADA_HOME=`command -v gnat`
-ADA_HOME="$(dirname "$ADA_HOME")/.."
+ADA_HOME=$(cd "$(dirname ${ADA_HOME})/.." 2>/dev/null && pwd)
 PLATFORM=`${ADA_HOME}/bin/gcc -dumpmachine`
 DEF_ADA_INCLUDE_PATH="${ADA_HOME}/lib/gcc/${PLATFORM}"
 ADA_GCC_VERSION=`${ADA_HOME}/bin/gcc -dumpversion`
 DEF_ADA_INCLUDE_PATH="${DEF_ADA_INCLUDE_PATH}/${ADA_GCC_VERSION}/rts-native/adainclude:${ADA_HOME}/include"
 export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-$DEF_ADA_INCLUDE_PATH}"
 
-export GPR_PROJECT_PATH="${GPR_PROJECT_PATH:-/opt/gnat/lib/gnat}"
-if [ ! -d "$GPR_PROJECT_PATH" ]; then
-   echo >&2 "GPR project path does not exists!"
+if [ -n "$GPR_PROJECT_PATH" -a ! -d "$GPR_PROJECT_PATH" ]; then
+   echo >&2 "GPR project path environment variable has been set to:"
+   echo >&2 "    \"${GPR_PROJECT_PATH}\""
+   echo >&2 "but that path is not a directory."
+   echo >&2 "Please set the environment variable GPR_PROJECT_PATH to the"
+   echo >&2 "location of the gnat libraries"
    exit 6
+else
+   # GPR_PROJECT_PATH not previously specified, use a default
+   export GPR_PROJECT_PATH="${ADA_HOME}/lib/gnat"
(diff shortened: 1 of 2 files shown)


experiments/list_unsupported.sh
```diff
@@ -1,6 +1,11 @@
 #!/bin/sh
+
+# gnat2goto on the path?
 if ! command -v gnat2goto > /dev/null; then
   echo >&2 "gnat2goto not on PATH!"
+  gnat2goto_bin=$(cd "$(dirname ${0})/../gnat2goto/install/bin" 2>/dev/null && pwd)
```
hannes-steffenhagen-diffblue: Really minor nitpick: This'll break if some crazy person decides to symlink this script. I'd expect a person doing this would also know how to use it already, so wouldn't need this help message, so I don't consider this an issue.

experiments/list_unsupported.sh
```diff
@@ -40,11 +40,27 @@ done
 
 echo "$1: Unsupported features\n" > "$file_name".txt
 
+# gnat on the path?
 if ! command -v gnat > /dev/null; then
    echo >&2 "Gnat not on PATH!"
+   # Take a wild guess at where gnat might be installed...
```
hannes-steffenhagen-diffblue: instead of doing this, how about checking if `locate` is on the path and using that to (maybe) find gnat if it is?

experiments/list_unsupported.sh
```diff
@@ -61,19 +61,33 @@ if ! (gnat --version | grep -q 'GNAT GPL 2016') ; then
    exit 5
 fi
 
+# Sane ADA_INCLUDE_PATH and GPR_PROJECT_PATH?
 GNAT2GOTO=`command -v gnat2goto`
 ADA_HOME=`command -v gnat`
-ADA_HOME="$(dirname "$ADA_HOME")/.."
+ADA_HOME=$(cd "$(dirname ${ADA_HOME})/.." 2>/dev/null && pwd)
```
hannes-steffenhagen-diffblue: does curly bracket expansion work if there are spaces on the path?

Also, I'm not sure, but is the `cd` just there to check if the path exists? If so, why not just `[ -d ... ]`?

experiments/list_unsupported.sh
```diff
@@ -61,19 +61,33 @@ if ! (gnat --version | grep -q 'GNAT GPL 2016') ; then
...
 PLATFORM=`${ADA_HOME}/bin/gcc -dumpmachine`
 DEF_ADA_INCLUDE_PATH="${ADA_HOME}/lib/gcc/${PLATFORM}"
 ADA_GCC_VERSION=`${ADA_HOME}/bin/gcc -dumpversion`
 DEF_ADA_INCLUDE_PATH="${DEF_ADA_INCLUDE_PATH}/${ADA_GCC_VERSION}/rts-native/adainclude:${ADA_HOME}/include"
 export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-$DEF_ADA_INCLUDE_PATH}"
 
-export GPR_PROJECT_PATH="${GPR_PROJECT_PATH:-/opt/gnat/lib/gnat}"
-if [ ! -d "$GPR_PROJECT_PATH" ]; then
-   echo >&2 "GPR project path does not exists!"
+if [ -n "$GPR_PROJECT_PATH" -a ! -d "$GPR_PROJECT_PATH" ]; then
```
hannes-steffenhagen-diffblue: FYI  `GPR_PROJECT_PATH` like many other path variables doesn't necessarily consist of only a single path, so this check actually excludes some possibly valid configurations. Not sure if that's ever going to come up, but it'll be a nuisance if it does.

experiments/list_unsupported.sh
```diff
@@ -88,7 +87,21 @@ else
       exit 9
    fi
 fi
+
+# Can we build the support tool?
+# Problem: GNAT 2016 helpfully installs a g++ binary alongside gnat... except
```
hannes-steffenhagen-diffblue: Yep, because `gnat` is actually just part of gcc.

experiments/list_unsupported.sh
```diff
@@ -88,7 +87,21 @@ else
...
+# it's way out of date so invoking g++ via PATH looking will pickup that gnat g++
```
hannes-steffenhagen-diffblue: FWIW we should probably also update our gnat to a newer release at some point

experiments/list_unsupported.sh
```diff
@@ -88,7 +87,21 @@ else
...
+# compiler, rather than the default system compiler... Instead we need to
+# temporarily drop the GNAT tools off the path while we build the tool.
+saved_path="$PATH"
+export PATH=$(echo ${PATH} | tr ':' '\n' | grep -v "${ADA_HOME}" | paste -s -d : - )
```
hannes-steffenhagen-diffblue: I think it'd be easier to search for system g++ via `command` before setting the PATH instead, and then just using the absolute path for this one command

experiments/list_unsupported.sh
```diff
@@ -1,5 +1,16 @@
...
 
+# Usage info
+usage()
+{
+   echo "Run GNAT2Goto on an Ada repository.\n"
+   echo "The output is an ordered list of currently unsupported features"
+   echo "with the number of times they occur in the input repository.\n"
+   echo "The script builds a parsing program using collect_unsupported.cpp and expects"
+   echo "this file to be in the same folder.\n"
+   echo "Usage:\n\nlist_unsupported.sh path_to_ada_source_folder"
```
hannes-steffenhagen-diffblue: I think this should probably be the first line?

experiments/list_unsupported.sh
martin-cs: Yeah.  `list_unsupported.sh` is intended as a development utility, not a final deliverable so this kind of robustness, although a good thought, is probably more than we need.

experiments/list_unsupported.sh
martin-cs: Good idea but I worry that it would find the system GNAT rather than the one we want (the 2016 one).  I know Trevor and I both have multiple versions of GNAT installed.

experiments/list_unsupported.sh
martin-cs: Yes although as I understand it this is a non-trivial job because there is a dependency between the version of GNAT that gnat2goto is based on and the version we use and 'rebasing' gnat2goto may well be non-trivial.  I am inclined to say it is not a task for this project.

experiments/list_unsupported.sh
hannes-steffenhagen-diffblue: I didn't mean to suggest that this should be used automatically as the path, just as a guide to give a more targeted suggestion along the lines of "hey, did you mean to add <this path here> to your PATH?"

experiments/list_unsupported.sh
chrisr-diffblue: Yeah, good catch - but I think this is good enough for now.

experiments/list_unsupported.sh
chrisr-diffblue: `locate` is also not posix standard I think (so it might not be on the target system, or if it is, might not be enabled - e.g. OSX has the locate command, but doesn't run the indexer). My main purpose here is to help someone who is following the documented install process, which recommends this location for gnat.

experiments/list_unsupported.sh
chrisr-diffblue: No, this is for path canonicalisation - it's crude, but idiomatic way to do so - `cd` into the directory, then `pwd` from that directory - it turns something like `foo/..` into `foo`, or `foo/bar/../wibble` into `foo/wibble`.

experiments/list_unsupported.sh
chrisr-diffblue: Ah, good point - yes, I'll change that.

experiments/list_unsupported.sh
chrisr-diffblue: The problem is that PATH is set outside this script, not within the script, hence this attempt to try to remove the gnat path from PATH.

experiments/list_unsupported.sh
chrisr-diffblue: Ah yes - I'd thought about this in an earlier cleanup but then did nothing about it. Again, for the moment I think the assumption of a single directory is OK for the purpose we have in mind.

experiments/list_unsupported.sh
NlightNFotis: Can this be turned into a function in a later PR? This sounds like useful information that will be lost otherwise. If we had a function called `path_canon` or something it would be easier to understand and use.

PR #256, opened 2019-06-14
Fix PR #220 Regression - Urgent

tjj2017:
PR #220  introduced a regression.  It entered subprogram parameters into
the symbol table before processing their declaration causing the
processing of a subprogram specification to fail.

The fault was found by the regression caused to PR #245 but could could happen if a subprogram body was declared without a separate declaration (perfectly legal to do this in Ada).

The problem of symbols missing from the symbol that PR #220 was designed to fix was due to not processing a subprogram's specification prior to processing its body, if it had no separate declaration.
 
This change removes the code from PR 220 and ensures a
subprogram specification is processed before its body.

I have run the regression tests on this branch and the results as for the main branch and as best I can tell the golden results are not significantly different, the summaries being identical.


tjj2017:
I have updated the golden results

PR #257, opened 2019-06-19
Add partial support for Rational APEX file naming conventions to list_unsupported.sh

chrisr-diffblue:
These changes make the script work OK with .1.ada and .2.ada file naming
but at the moment gnat2goto still expects to import from .ads files.
We need to switch gnat2goto mode somehow when --apex is specified to this
script.

### DIFF
--- a/experiments/list_unsupported.sh
+++ b/experiments/list_unsupported.sh
@@ -3,13 +3,21 @@
 # Usage info
 usage()
 {
-   echo "Usage:\n\nlist_unsupported.sh path_to_ada_source_folder\n"
-   echo "Run GNAT2Goto on an Ada repository.\n"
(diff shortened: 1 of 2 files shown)


chrisr-diffblue:
@martin-cs - Would you like to review this now? I've made the script add a `.adc` file to the gnat2goto command line to handle APEX naming conventions - running this script with `--apex` should now produce a report for these files. In addition, I've also added a `--adc` option to allow a user supplied `.adc` file to be specified instead. I have experimented with using `gnatname` to generate an appropriate `.adc` file, but at the moment I cannot get it recognise any of the files it finds as units, so it just generates an empty `.adc` file.

martin-cs:
I have tested this as best as I can without access to a current PTC ApexAda code-base.

PR #258, opened 2019-06-22
Subtype constraint

tjj2017:
This change handles array subtypes and declarations constrained by subtype marks and anonymous subtypes.

The changes are related to the subprogram Make_Array_Default_Initialiser.
Note: The subprogram has been changed to handle subtype constraints but not changed to handle multi-dimensional arrays.  The original subprogram did not handle multi-dimensional arrays.  I do not know if that is a deliberate decision to exclude multi-dimensional arrays from the handled features or an oversight?

I can see how to handle multi-dimensional arrays from the front-end perspective but have not tried to understand how they would be modelled in GOTO.

### DIFF
--- a/experiments/golden-results/SPARK-tetris-summary.txt
+++ b/experiments/golden-results/SPARK-tetris-summary.txt
@@ -3,6 +3,11 @@ Calling function: Process_Declaration
 Error message: Package declaration
 Nkind: N_Package_Declaration
 --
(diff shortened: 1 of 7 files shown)


tjj2017:
I now have a Linux virtual machine and I have updated the golden-results.

More text is now parsed and this has introduced extra 0r changed errors.

PR #260, opened 2019-06-25
Address clause

tjj2017:
gnat2goto is crashing when an address representation clause is encountered.  The reason is that gnat2goto was assuming that the expression defining the attribute for all representation clauses is an integer literal.  This is certainly not the case for an address clause because a subprogram has to be called to convert the given integer value to a private Address type.

This mod separates out address clauses before the expression is evaluated and reports a missing feature.  No doubt we will have to deal with address clauses in the future when we decide how to represent and perform some sort of analysis on them.

Assuming that the expression defining the value of a Size or Component_Size attribute is an integer_literal is also too restrictive, it can be a static_expression.  I have replaced IntVal which only accepts an integer_literal with Expr_Value which accepts a static_expression and returns its folded value.

I have added a successful test which defined modular types with a Size representation clause with static_expressions.

I have also added a failing test which legally applies a size representation clause to an object but fails.  In addition, size representation clauses applied to derived integer types, again legal in Ada, fails in gnat2goto.

The reason for these failures is the unsatisfied assertion:
                  --  Just check that the front-end already applied this size
                  --  clause, i .e. that the size of type-irep we already had
                  --  equals the entity type this clause is applied to (and the
                  --  size specified in this clause).
                  pragma Assert
                    (Entity_Esize =
                       UI_From_Int (Int (Get_Width (Target_Type_Irep)))
                     and Entity_Esize = Expression_Value);
I have not tried to fix this as I am unsure of the GOTO representation that is being used.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -1,6 +1,7 @@
 with Namet;                 use Namet;
 with Nlists;                use Nlists;
 with Sem;
+with Sem_Eval;              use Sem_Eval;
 with Sem_Util;              use Sem_Util;
 with Sem_Aux;               use Sem_Aux;
 with Snames;                use Snames;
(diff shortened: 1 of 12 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4644,51 +4645,70 @@ package body Tree_Walk is
...
-                 Follow_Symbol_Type (Get_Subtype (Get_Type (Array_Data)),
-                                     Global_Symbol_Table);
-               Target_Subtype_Width : constant Uint :=
-                 UI_From_Int (Int (Get_Width (Target_Subtype)));
+               Target_Name : constant Irep := Do_Identifier (Name (N));
+               Entity_Esize : constant Uint := Esize (Entity (N));
+               Target_Type_Irep : constant Irep :=
+                 Follow_Symbol_Type
+                   (Get_Type (Target_Name), Global_Symbol_Table);
+               Expression_Value : constant Uint := Expr_Value (Expression (N));
```
xbauch: Effectively, this line is what is changed in this commit. And it seems ok.

tjj2017:
In regression tests a different node number is given in test address_clause between Mac and Linux.  I have changed the expected test.out file to that for Linux.  Probably should avoid having node numbers in test.out.

Updated golden-results

martin-cs:
( Please make sure this is synchronised with https://github.com/diffblue/gnat2goto/pull/233 )

xbauch:
> ( Please make sure this is synchronised with #233 )

As far as I understand, this PR generalises handling the other two representation clauses (`size` and `component_size`). However we choose to implement `address` (via #233 or otherwise) will be possible to plug in next to this PR. So yes, we're in sync.

tjj2017:
Rebased to current diffblue/master and updated golden-results

PR #262, opened 2019-06-29
Add test for enumeration out parameters

tjj2017:
Test currently fails with gnat2goto raising:
SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed precondition from ireps.ads:1875

gnat2goto fails with above precondition failure when translating UKNI/main.adb line 247 which calls a procedure with an enumeration out parameter.  This small test program replicates the same fault.

The problem seems to lie in gnat2goto driver/tree_walk.adb procedure Handle_Parameter.  The function Wrap_Argument is called with a parameter obtained from the call of the function Handle_Enum_Symbol_Members which appears to return the wrong sort of Irep for the function Make_Address_Of called by Wrap_Argument.

### DIFF
--- a/testsuite/gnat2goto/tests/enum_out_param/enum_out.adb
+++ b/testsuite/gnat2goto/tests/enum_out_param/enum_out.adb
@@ -1,0 +1,14 @@
+procedure Enum_Out is
+   type Enum is (One, Two, Three);
+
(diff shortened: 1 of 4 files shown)


tjj2017:
This problem has been raised as Issue #264

PR #266, opened 2019-07-29
Implement handling N_In

xbauch:
That is the check that a value `x` is inside a range `<low,high>`. The implementation simply builds a conjunction `x \geq low /\ x \leq high`. Re-enables existing test (changed so that the assertion is failing).

### DIFF
--- a/experiments/golden-results/UKNI-Information-Barrier-summary.txt
+++ b/experiments/golden-results/UKNI-Information-Barrier-summary.txt
@@ -1,4 +1,4 @@
-Occurs: 47 times
+Occurs: 46 times
 Calling function: Do_Expression
 Error message: Unknown expression kind
 Nkind: N_Expanded_Name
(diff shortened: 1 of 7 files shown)


experiments/golden-results/UKNI-Information-Barrier-summary.txt
```diff
@@ -1,13 +1,8 @@
-Occurs: 47 times
+Occurs: 46 times
```
martin-cs: I'm obviously happy to see these going down but do you see why this change reduces the occurrences of this error?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1419,6 +1417,28 @@ package body Tree_Walk is
       end case;
    end Do_Expression;
 
+   function Do_In (N : Node_Id) return Irep is
+      Left_Op : constant Irep := Do_Expression (Left_Opnd (N));
```
martin-cs: Does GNAT guarantee that this node will have lower and upper bounds like this?  For example with subtype inclusion?

xbauch: I'm not entirely sure. I've added a check (to `Do_Expression`) that reports as unsupported feature if we ended up calling `Do_In` with a node that doesn't have the bounds.

PR #267, opened 2019-07-29
Implement looping over user-type

xbauch:
In case that type is a range type. For example `for x in range_type` where `range_type = from
.. to`. Includes a test.

### DIFF
--- a/testsuite/gnat2goto/tests/loop_in_custom_type/test.adb
+++ b/testsuite/gnat2goto/tests/loop_in_custom_type/test.adb
@@ -1,0 +1,12 @@
+procedure Test is
+   From : constant := 1;
+   To : constant := 5;
+   subtype My_Type is Integer range From .. To;
+   Iter_Counter : Integer := 0;
+begin
+   for X in My_Type loop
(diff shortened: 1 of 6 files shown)


testsuite/gnat2goto/tests/loop_in_custom_type/test.adb
```diff
@@ -0,0 +1,9 @@
+procedure Test is
+   From : constant := 1;
+   To : constant := 5;
+   subtype My_Type is Integer range From .. To;
+begin
+   for X in My_Type loop
+      pragma Assert (X>4);
+   end loop;
```
chrisr-diffblue: Is there any value in adding some assertions that we also expect to succeed? For instance, perhaps having another loop that maybe just counts iterations, followed by an assertion outside the loop that asserts the counter value?

xbauch: Done.

chrisr-diffblue: Thanks :-)

PR #268, opened 2019-08-02
Expanded name

tjj2017:
In Do_Expression N_Expanded_Name was not being handled which led to many faults including rating an exception due to a failed precondition at arrays.ads line 57.

The gnat front-end replaces all instances of a selected_component by an expanded_name except for those instances which are actually a component selection.

In Do_Expression N_Expanded_Name is treated as an N_Identifier.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -149,6 +149,11 @@ Error message: Unsupported pragma: No return
 Nkind: N_Pragma
 --
 Occurs: 5 times
+Calling function: Do_Expression
+Error message: Unknown expression kind
+Nkind: N_Expanded_Name
+--
+Occurs: 5 times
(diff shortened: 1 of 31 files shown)


tjj2017:
Updated golden results again.  I don't know what went wrong the first time?

xbauch:
@tjj2017 it seems you have had an old version of `master` when committing this work. I guess the easiest fix would be to `git remote update diffblue --prune` start a new local branch based on `diffblue:master`, cherry-pick the one commit from here, run and update the golden results, and then force-push into this branch.

chrisr-diffblue:
@tjj2017 - I think this just needs rebasing to update the golden results, then we should be able to merge.

tjj2017:
I have been unable to resolve the conflicts remotely and have done it directly via GitHub.

xbauch:
Merged as part of #271.

PR #272, opened 2019-08-15
Fix enum in param

xbauch:
By casting each occurence of `I_C_Enum` to `int`. The first commit is a squash of prerequisites -- no need to review it here.


xbauch:
@martin-cs This fix originates from https://github.com/martin-cs/ASVAT/issues/81, the changes in the golden results were not intentional -- I guess we just got a bit further down the ASTs.

martin-cs:
@xbauch OK... but the number of unsupported features has gone down.  This is not, in of itself, a problem.  I mean, in fact, it's good but I just wanted to make sure that you had some idea of /why/ this was resolving things.

martin-cs:
Thanks for the rebase, the changes to the golden results now look reasonable.

PR #273, opened 2019-08-15
Low up bound

tjj2017:
Modified code now handles discrete_subtype_definitions that are not a range.

A function, Get_Range has been introduced to wrap calls to Low/High_Bound to
ensure that these are always called with a range node.


tjj2017:
I have missed attribute_reference from possible discrete_range_type.  From the few tests I did, this seemed to be resolved by the front end but is not always the case. Please do not review this PR until I have fixed this problem

tjj2017:
Work in progress

PR #277, opened 2019-08-26
Subtype constraint

tjj2017:


tjj2017:
Unfortunately fixing the problems of range constraint exceptions are more widespread than I thought.  I will have to come back to this after the meeting.

xbauch:
Merged as part of #283.

PR #279, opened 2019-08-27
Fix "mod of unsupported type" error

hannes-steffenhagen-diffblue:
Sometimes we'd get mod operations on mod types because they collapse to
unsignedbv. This adds a case handler for this.

No test case because while I can observe this happening in our projects I wasn't able to find a small reproducer.

### DIFF
--- a/experiments/golden-results/UKNI-Information-Barrier-summary.txt
+++ b/experiments/golden-results/UKNI-Information-Barrier-summary.txt
@@ -38,11 +38,6 @@ Calling function: Do_Expression
 Error message: Unknown expression kind
 Nkind: N_Range
 --
-Occurs: 1 times
-Calling function: Do_Operator_General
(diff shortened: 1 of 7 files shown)


experiments/golden-results/UKNI-Information-Barrier-summary.txt
```diff
@@ -39,9 +39,9 @@ Error message: Unknown expression kind
 Nkind: N_Range
 --
 Occurs: 1 times
-Calling function: Do_Operator_General
-Error message: Mod of unsupported type
-Nkind: N_Op_Not
+Calling function: Process_Declaration
+Error message: Use package clause declaration
+Nkind: N_Use_Package_Clause
```
hannes-steffenhagen-diffblue: I'm not sure why this is here... there's no `N_Use_Package_Clause` in my golden results?

PR #281, opened 2019-08-28
Process sub-packages

xbauch:
during process declarations, using the same procedure we already had for
package-declarations. Includes a test.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -8,15 +8,10 @@ Calling function: Process_Declaration
 Error message: Generic declaration
 Nkind: N_Generic_Subprogram_Declaration
 --
-Occurs: 76 times
+Occurs: 78 times
 Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: Postcondition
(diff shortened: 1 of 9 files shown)


experiments/golden-results/StratoX-summary.txt
```diff
@@ -43,56 +43,51 @@ Calling function: Do_Withed_Unit_Spec
...
+Occurs: 19 times
 Calling function: Process_Declaration
 Error message: Generic instantiation declaration
 Nkind: N_Function_Instantiation
 --
-Occurs: 11 times
+Occurs: 15 times
+Calling function: Process_Statement
+Error message: Unknown expression kind
+Nkind: N_Object_Declaration
```
martin-cs: I recall seeing a patch that resolved an error that looked like this on UKNI.  I'm not sure who wrote it; I think it was either you or @hannes-steffenhagen-diffblue .  If that has been merged, should we be seeing these?

chrisr-diffblue: This doesn't look like this has been *introduced* by this PR, it's just moved in the file because of other changes in the missing feature list.

PR #283, opened 2019-09-09
Rework of subtype constraint

xbauch:
Note that commits 5,6 are from prerequisite PR #280 -- no need to review them here. On top of #277 I only added a simple case analysis for mod-subtype bounds (7th commit). Now there are no constraint-related issues in UKNI.

PR #284, opened 2019-09-16
Explicit ada checks

xbauch:
In preparation for better handling of the ada-checks we replace the existing range-check with `__CPROVER_Ada_Range_Check` and create similar checks for div-by-zero and overflow.

The first commit is a squash of the prerequisite PR #283 -- no need to review it here.


gnat2goto/driver/goto_utils.adb
```diff
@@ -18,6 +18,93 @@ package body GOTO_Utils is
       return Size_T;
    end CProver_Size_T;
 
+   Float_T : Irep := Ireps.Empty;
```
hannes-steffenhagen-diffblue: These are all a bit misleading as the C standard doesn't really guarantee any of these sizes and it's not like we're getting them from an actual C compiler

gnat2goto/driver/goto_utils.adb
```diff
@@ -532,11 +619,108 @@ package body GOTO_Utils is
...
          return Expr;
       end if;
    end Cast_Enum;
+
+   function Make_Assert_Call (Assertion : Irep;
+                              Description : Irep; Source_Loc : Source_Ptr)
+                              return Irep is
+      SE_Call_Expr : constant Irep :=
+        New_Irep (I_Code_Function_Call);
+      Sym_Assert   : constant Irep := New_Irep (I_Symbol_Expr);
```
hannes-steffenhagen-diffblue: AAAAAHHHHHH

gnat2goto/driver/goto_utils.adb
```diff
@@ -532,11 +619,108 @@ package body GOTO_Utils is
...
       else
          return Expr;
       end if;
    end Cast_Enum;
+
+   function Make_Assert_Call (Assertion : Irep;
+                              Description : Irep; Source_Loc : Source_Ptr)
+                              return Irep is
+      SE_Call_Expr : constant Irep :=
+        New_Irep (I_Code_Function_Call);
```
hannes-steffenhagen-diffblue: AHHHHHHHHHH

gnat2goto/driver/goto_utils.adb
```diff
@@ -532,11 +619,108 @@ package body GOTO_Utils is
...
+      Sym_Assert   : constant Irep := New_Irep (I_Symbol_Expr);
+      Assert_Args  : constant Irep := New_Irep (I_Argument_List);
```
hannes-steffenhagen-diffblue: AHHHHHHHHHH

gnat2goto/driver/goto_utils.adb
```diff
@@ -532,11 +619,108 @@ package body GOTO_Utils is
...
+   begin
+      Set_Identifier (Sym_Assert, "__CPROVER_assert");
```
hannes-steffenhagen-diffblue: Could you replace those with regular constructor calls?

gnat2goto/driver/tree_walk.adb
```diff
@@ -3506,17 +3506,28 @@ package body Tree_Walk is
                Global_Symbol_Table));
       Set_Type (Ret, Ret_Type);
 
-      if Do_Overflow_Check (N) then
-         Set_Overflow_Check (Ret, True);
-      end if;
+      declare
+         Maybe_Overflow_Check : Irep := Ret;
+         Maybe_Division_Check : Irep := Ret;
```
hannes-steffenhagen-diffblue: ? Why are these both the same?

NVM, I see it now :(((

testsuite/gnat2goto/lib/python/test_support.py
```diff
@@ -79,10 +79,12 @@ def process_cbmc(debug, files, cbmcargs):
     cbmcerr_file = open (cbmcerr)
     cbmcerr_text = cbmcerr_file.read()
     cbmcerr_file.close()
+    CBMC_VER_SUCC_CODE=0
+    CBMC_VER_FAIL_CODE=10
```
hannes-steffenhagen-diffblue: :+1:

gnat2goto/driver/range_check.adb
```diff
@@ -144,20 +160,53 @@ package body Range_Check is
...
-   -- Make_Range_Assert_Expr --
-   ----------------------------
+
+   function Make_Overflow_Assert_Expr (N : Node_Id; Value : Irep) return Irep
+   is
+      Value_Type : constant Irep :=
+        Follow_Symbol_Type (Get_Type (Value), Global_Symbol_Table);
+      Low_Value : constant Irep := Get_Bound (N, Value_Type, Bound_Low);
+      High_Value : constant Irep := Get_Bound (N, Value_Type, Bound_High);
+      Maybe_Casted_Value : constant Irep := Value;
```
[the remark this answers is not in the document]
xbauch: I need to cast `value` to the type of `low_value`. But if I casted the whole expression it would be evaluated in the wider type during the assertion function call. That's why I cast the left-right hand side of the binary expression separately. I will add a case for unary expressions (didn't think they can overflow).

gnat2goto/driver/range_check.adb
```diff
@@ -108,12 +108,28 @@ package body Range_Check is
...
+               return Report_Unhandled_Node_Irep (N, "Get_Bound",
+                                                  "Unsupported float width");
+            end if;
+            return Make_Constant_Expr (Source_Location => No_Location,
+                                       I_Type          => C_Double_T,
+                                       Range_Check     => False,
+                                       Value           =>
+                                         (if Pos = Bound_Low
+                                          --  0xff7fffff as double
+                                          then "C7EFFFFFE54DAFF8"
```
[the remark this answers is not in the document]
xbauch: Names given and width described. I'm basically forming an expression `value >= smallest_float` so I though that if `value` was -Infty the expression would evaluate to false, etc.

gnat2goto/driver/range_check.adb
```diff
@@ -108,12 +108,28 @@ package body Range_Check is
...
+                                          I_Type          =>
+                                          Maybe_Double_Type_Width (Bound_Type),
                                           Range_Check     => False,
                                           Value =>
                                             Convert_Uint_To_Hex (Bound_Value,
-                                              Types.Pos (Type_Width)));
+                                              Types.Pos (Type_Width * 2)));
             end;
+         when I_Floatbv_Type =>
+            if Get_Width (Bound_Type) /= 32 and Get_Width (Bound_Type) /= 64
```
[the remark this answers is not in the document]
xbauch: Well not for this, but I tried to use it anyway.

gnat2goto/driver/goto_utils.adb
xbauch: Done: there should be no `new_irep` in the added code.

xbauch:
@martin-cs Hi, Martin. This PR is supposed to be the first part of https://github.com/martin-cs/ASVAT/issues/9. What do you think?

experiments/golden-results/vct-summary.txt
```diff
@@ -300,15 +300,10 @@ Error detected at REDACTED
...
+| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:4840                     |
 Error detected at REDACTED
 --
 Occurs: 1 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:4829                     |
+| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:4840                     |
 Error detected at REDACTED
---
-Occurs: 2 times
```
martin-cs: This is good.

gnat2goto/driver/goto_utils.adb
martin-cs: C standard doesn't but IEEE-754 does and there is a link between the Ada standard and it.

martin-cs: Ah, scratch that.  Yes, it's less clear what is going on and why this is needed.  ISO 9899 gives minimum ranges for each type but not their actual width.  I would be more comfortable, if you need C compatible things, with using `int64_t` et al.

gnat2goto/driver/range_check.adb
```diff
@@ -108,11 +108,31 @@ package body Range_Check is
...
+                                          Maybe_Double_Type_Width (Bound_Type),
                                           Range_Check     => False,
                                           Value =>
                                             Convert_Uint_To_Hex (Bound_Value,
-                                              Types.Pos (Type_Width)));
+                                              Types.Pos (Type_Width * 2)));
+            end;
+         when I_Floatbv_Type =>
+            declare
+               Smallest_Float : constant String := "C7EFFFFFE54DAFF8";
```
martin-cs: What is the relationship between this string and the comment below?

gnat2goto/driver/range_check.adb
```diff
@@ -108,11 +108,31 @@ package body Range_Check is
...
+               --  0xff7fffff : single-precision min stored in double precision
+               Largest_Float : constant String := "47EFFFFFE54DAFF8";
```
martin-cs: Likewise.

gnat2goto/driver/range_check.adb
```diff
@@ -144,20 +164,160 @@ package body Range_Check is
...
+
+      ---------------------------
+      -- Build_Assert_Function --
+      ---------------------------
+
+      --  Build a symbol for the following function
+      --  Value_Type division_check(Value_Type value, Value_Type divisor)
+      --  {
+      --    assert (divisor != 0);
+      --    return value;
```
martin-cs: A couple of things : 1. the default behaviour of all of the check functions should be `assert(condition); assume(condition)`, 2. why pass `Value_Type`?  It should be enough to take a single Boolean?

gnat2goto/driver/goto_utils.ads
```diff
@@ -18,6 +18,15 @@ package GOTO_Utils is
    --  TODO change this to Irep
    function Internal_Source_Location return Source_Ptr;
 
+   function C_Float_T return Irep;
+   function C_Double_T return Irep;
+   function C_Int_T return Irep;
+   function C_Long_T return Irep;
```
martin-cs: What are these for?

gnat2goto/driver/range_check.adb
xbauch: Equivalence. The number `-3.40282347E+39` is the small number representable in single-precision floats, and it's hex is `0xff7fffff`. The same number in double precision is `0xC7EFFFFFE54DAFF8`.

gnat2goto/driver/goto_utils.adb
xbauch: renamed.

gnat2goto/driver/range_check.adb
xbauch: Ad 1. The assert-then-assume behaviour is there, just the comment was wrong -> Updated. Ad 2. the division (and other) check may be call inside expression, thus it needs to return the result value of the division.

gnat2goto/driver/goto_utils.ads
xbauch: Shortcuts, so that we don't create a new irep every time we refer to the integer type. Also it helps implementing the `Maybe_Double_Type_Width` function.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -1,3 +1,8 @@
+Occurs: 212 times
```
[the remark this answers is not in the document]
xbauch: The precondition for `Get_Bound` was relaxed. That alone probably does not account for all the new unsupported postcondition-pragmas but at least when analysing `a-ngcefu.adb` in Stratox, gnat2goto used to give up on the `Get_Bound` precondition whereas now it carries on.

xbauch:
@martin-cs What do you think now?

gnat2goto/driver/range_check.adb
martin-cs: Ummm... how did you come by that number?  The low bytes look a bit wonky to me.  Something I learned at great cost some years ago is that decimal is not a good way of working with binary floating-point.  It should only be used as the final step and for presentation.  Don't store floats as decimal and hope to get them back exactly, esp. crossing formats.

The most negative normal number in single precision floating-point is `-0x1.fffffep+127f` which has the bit pattern `0xff7fffff` and decimal value `-3.402823e+38f`.

The number in double is `-0x1.fffffep+127` (the last character is significant in C!) which then has bit pattern `0xc7efffffe0000000` and decimal value `-3.402823e+38`.

gnat2goto/driver/goto_utils.ads
martin-cs: Fair.

experiments/golden-results/StratoX-summary.txt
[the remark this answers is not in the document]
martin-cs: Seems legit.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -28,17 +28,27 @@ Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: No strict aliasing
 Nkind: N_Pragma
 --
-Occurs: 30 times
+Occurs: 31 times
```
martin-cs: I presume this is another "things continue further so they hit later errors"?

If this is what is going on, shouldn't the total number of errors stay the same?  Does that happen?

experiments/golden-results/vct-summary.txt
martin-cs: Still like this...

gnat2goto/driver/goto_utils.adb
martin-cs: Thank you.

gnat2goto/driver/goto_utils.adb
```diff
@@ -549,11 +674,146 @@ package body GOTO_Utils is
...
+
+      return Make_Code_Function_Call (Arguments       => Assume_Args,
+                                      I_Function      => Sym_Assume,
+                                      Lhs             => Make_Nil (Source_Loc),
+                                      Source_Location => Source_Loc,
+                                      I_Type          => Make_Void_Type,
+                                      Range_Check     => False);
+   end Make_Assume_Call;
+
+   function Get_Integer_Zero return Irep
```
martin-cs: `Get_Int32_T_Zero`?

gnat2goto/driver/range_check.adb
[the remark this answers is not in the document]
martin-cs: Unary negate can overflow...

martin-cs:
> @martin-cs What do you think now?

@xbauch I think I shouldn't review things at the start of term!  Much better but still some stuff to do.

gnat2goto/driver/range_check.adb
xbauch: I took the `0xff7fffff`, converted that to as 32-bit float to decimal (as above, but the power is 38, that's my typo). Then I converted that decimal to 64-bit double (via some online converter) and got `0xC7EFFFFFE54DAFF8`. This really is not my area so I can use your numbers if you prefer.

martin-cs: That sounds consistent with what result you have got.  If you want the largest negative single precision float, as a double, then it is `0xc7efffffe0000000`.

xbauch: Ok, so I assume the largest positive single float as double is `0x47EFFFFFE0000000`?

xbauch:
@martin-cs the last-but-one commit addresses the things you asked for (mainly the minus overflow check).

gnat2goto/driver/range_check.adb
martin-cs: Yes!

gnat2goto/driver/range_check.adb
martin-cs: OK... this may complicate what the "ignore" option is for these.

testsuite/gnat2goto/tests/constant_declarations/test.out
```diff
@@ -1,2 +1,7 @@
+Error from cbmc test:
+function `__CPROVER_Ada_Overflow_Check' in module `' is shadowed by a definition in module `'
+function `__range_check5' in module `' is shadowed by a definition in module `'
+
```
martin-cs: ???

testsuite/gnat2goto/tests/int_overflow/test.py
```diff
@@ -1,4 +1,3 @@
 from test_support import *
-    
-prove("--signed-overflow-check")
```
martin-cs: Progress!

testsuite/gnat2goto/tests/constant_declarations/test.out
xbauch: It's a warning from `linking.cpp`: when we try to merge the two symbol tables they both contain functions with the same name. And if their types match, CBMC only emits this warning and throws away the _new_ definition. Which is ok for `__CPROVER_Ada_Overflow_Check` but maybe not for the `__range_checkn` which might be unique. I think we should mangle the file-name into the range-check function name somehow.

martin-cs: I'm not sure I understand, where is the `__range_checkn` function coming from?  If we are generating it why isn't it `__CPROVER_Ada_Range_Check` or similar?

xbauch: Well `__range_check` is something you can use inside an expression so it has to evaluate to the same type as the expression you are range-checking. It then calls `__CPROVER_Ada_Check` which has type `void (int, char*)`.

martin-cs: Right; this is a little bit concerning because it suggests that when we link things we risk using the wrong range check.  Maybe the range check function should include the type so that they are unique.

( Minor point : maybe is should call `__CPROVER_Ada_Range_Check`)

PR #285, opened 2019-09-16
Refactor/remove New_Irep from driver.adb

hannes-steffenhagen-diffblue:
`New_Irep`, and generally using the `Set_*` stuff is a nuisance for maintainability, as any change to the allowed and required fields will not be noticed at compile time. This is the first PR among a series of PRs with the goal of purging it from the codebase entirely and eventually removing it.

This is also preparatory work to fix a long standing issue of assertions just having their assertion number and source locations, but no text or human readable condition.

Also adds/changes some utility functions (particularly float format related stuff now gets passed the width of the floating point type directly rather than an Irep which we only use to get the width).

### DIFF
--- a/gnat2goto/driver/driver.adb
+++ b/gnat2goto/driver/driver.adb
@@ -416,11 +416,6 @@ package body Driver is
                others => <>);
             Initial_Call_Args : constant Irep := Make_Argument_List;
             Entry_Procedure : constant Irep := Symbol_Expr (Program_Symbol);
-            Initial_Call      : constant Irep := Make_Code_Function_Call
-              (Arguments => Initial_Call_Args,
-               I_Function => Entry_Procedure,
-               Source_Location => No_Location);
-
             Program_Args : constant Irep_List :=
               Get_Parameter (Get_Parameters (Get_Type (Entry_Procedure)));
          begin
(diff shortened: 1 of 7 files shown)


gnat2goto/driver/driver.adb
```diff
@@ -435,59 +438,57 @@ package body Driver is
                      Arg_Type : constant Irep := Get_Type (Arg);
                      Arg_Id   : constant Symbol_Id :=
                        Intern ("input_" &  Get_Identifier (Arg));
-                     Arg_Symbol : Symbol;
+                     Arg_Symbol : constant Symbol :=
```
xbauch: There is a wrapper for creating function arguments in `goto_utils`: `Create_Fun_Parameter` and `Param_Symbol`.

gnat2goto/driver/driver.adb
```diff
@@ -435,59 +438,57 @@ package body Driver is
...
+                  Return_Symbol : constant Symbol :=
+                    (Name | BaseName | PrettyName => Return_Id,
+                     Mode => Intern ("C"),
+                     SymType => Get_Return_Type (Get_Type (Entry_Procedure)),
+                     IsLValue => True,
+                     IsStaticLifetime => True,
+                     others => <>);
+                  Return_Expr : constant Irep := Symbol_Expr (Return_Symbol);
+                  Return_Decl : constant Irep := Make_Code_Decl
+                    (Symbol => Return_Expr,
```
xbauch: Is the line 497 still necessary?

hannes-steffenhagen-diffblue: No, thanks, fixed

gnat2goto/driver/driver.adb
hannes-steffenhagen-diffblue: Hm, I think I prefer this still

PR #287, opened 2019-09-17
Add index check

xbauch:
when parsing the `n_indexed_component` node. Using the same methods as for the
range check, i.e. first_index < index < last_index. Includes a test.


martin-cs:
As #284 has been merged is there any chance of a rebase of this?

xbauch:
> Does this generate a different name of `__CPROVER_Ada_..._check` function as it would be good to be able to handle array and type range checks differently.

Yes it's `__CPROVER_Ada_Index_Check` here.

PR #288, opened 2019-09-20
Remove new irep range check

NlightNFotis:
Substitute references to `New_Irep` with the appropriately typed constructors.

### DIFF
--- a/gnat2goto/driver/range_check.adb
+++ b/gnat2goto/driver/range_check.adb
@@ -152,8 +152,7 @@ package body Range_Check is
                                     Bounds_Type : Irep)
                                     return Irep
    is
-      Call_Inst : constant Irep := New_Irep (I_Side_Effect_Expr_Function_Call);
-      Call_Args : constant Irep := New_Irep (I_Argument_List);
+      Call_Args : constant Irep := Make_Argument_List;
       Actual_Type : constant Irep := Follow_Symbol_Type (Get_Type (Value),
                                                          Global_Symbol_Table);
       Followed_Bound_Type : constant Irep := Follow_Symbol_Type (Bounds_Type,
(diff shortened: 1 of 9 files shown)


gnat2goto/driver/range_check.adb
```diff
@@ -267,12 +264,57 @@ package body Range_Check is
...
+         else
+            --  If the bounds are of smaller type then we cast the bounds
+            --  to the type of the value being checked
+            Adjusted_Lower_Bound :=
+              Typecast_If_Necessary (Lower_Bound, Value_Expr_Type,
+                                     Global_Symbol_Table);
+            Adjusted_Upper_Bound :=
+              Typecast_If_Necessary (Upper_Bound, Value_Expr_Type,
+                                     Global_Symbol_Table);
+            return True;
```
xbauch: The function returns `True` in both. If you returned `False` here, you'd only need to swap the cases on lines 309 and 314.

gnat2goto/driver/range_check.adb
```diff
@@ -267,12 +264,57 @@ package body Range_Check is
...
+         end if;
+      end Set_Value_And_Bounds;
+
+      Values_Adjusted : constant Boolean := Set_Value_And_Bounds
+         (Adjusted_Value_Expr, Adjusted_Lower_Bound, Adjusted_Upper_Bound);
+
+      Op_Geq : constant Irep := Make_Op_Geq (
+        Lhs => (if Values_Adjusted then Adjusted_Value_Expr else Value_Expr),
```
xbauch: But then again why not set lhs to `Adjusted_Value_Expr` (it's initialised to `value_expr`). Same for `rhs`. Then the function above could be a procedure.

NlightNFotis:
@hannes-steffenhagen-diffblue @xbauch I pushed up a new commit that has refactored the function to look more in line with what hannes had suggested. Can both of you take a look if it's okay now?

gnat2goto/driver/range_check.adb
```diff
@@ -267,12 +264,51 @@ package body Range_Check is
...
+           Lower_Bound : Irep;
+        end record;
+
+      function Get_Adjusted_Value_And_Bounds
+         return Adjusted_Value_And_Bounds_T;
+      function Get_Adjusted_Value_And_Bounds
+         return Adjusted_Value_And_Bounds_T
+      is
+         Greater_Width : constant Boolean :=
+          Get_Width (Bound_Type) > Get_Width (Value_Expr_Type);
```
hannes-steffenhagen-diffblue: :pick: In this case I think it'd actually look better with the condition being inlined

PR #289, opened 2019-09-23
Discriminated record formation is broken

xbauch:
the test is now marked as failing.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -3123,7 +3123,16 @@ package body Tree_Walk is
                      Variant_Disc_Value := Disc_Actual;
                   end if;
 
-                  New_Expr := Do_Expression (Disc_Actual);
+                  if Present (Disc_Actual) then
+                     New_Expr := Do_Expression (Disc_Actual);
+                  else
+                     --  Default initialize to 0
+                     New_Expr := Make_Constant_Expr
+                       (Source_Location => Get_Source_Location (E),
+                        I_Type          => Do_Type_Reference (Etype (Iter)),
+                        Range_Check     => False,
+                        Value           => "0");
+                  end if;
                   Append_Struct_Member (Ret, New_Expr);
                   --  Substitute uses of the discriminant in the record
                   --  initialiser for its actual value:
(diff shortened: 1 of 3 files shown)


xbauch:
The second commit proposes a quick fix.

gnat2goto/driver/tree_walk.adb
```diff
@@ -2878,7 +2878,16 @@ package body Tree_Walk is
...
-                  New_Expr := Do_Expression (Disc_Actual);
+                  if Present (Disc_Actual) then
+                     New_Expr := Do_Expression (Disc_Actual);
+                  else
+                     New_Expr :=
+                       Make_Constant_Expr (Source_Location => Sloc (E),
+                                           I_Type          =>
+                                             Do_Type_Reference (Etype (Iter)),
+                                           Range_Check     => False,
+                                           Value           => "0");
```
hannes-steffenhagen-diffblue: Looks like a dummy value? Shouldn't this be an unsupported feature report or an exception, something of the sort then?

gnat2goto/driver/tree_walk.adb
```diff
@@ -4083,12 +4095,15 @@ package body Tree_Walk is
    function Do_Selected_Component (N : Node_Id) return Irep is
       Root           : constant Irep := Do_Expression (Prefix (N));
       Component      : constant Entity_Id := Entity (Selector_Name (N));
+      Orig_Component : constant Entity_Id :=
```
hannes-steffenhagen-diffblue: Can you add a short explanation for what an original record component is and why we need it here?

gnat2goto/driver/tree_walk.adb
xbauch: Yes. Or better yet, we should find out how to properly initialise the struct-member so that we don't end up calling `Discriminant_Default_Value` on a node that doesn't have one.

gnat2goto/driver/tree_walk.adb
xbauch:
```
struct Foo { int a; };
struct Foo bar;
bar.a = 5;
```
So the `Unique_Name` of the struct-component in declaration is `Foo__a`. But when parsing the assignment the component is that of the entity `bar`, thus it's unique-name is `bar__a`: which is not present in `Foo`. That's why we have to use the component of the original-record to get the right name. (I'll add this as a comment.)

testsuite/gnat2goto/tests/discriminated_record/discriminated_record.adb
```diff
@@ -15,4 +15,6 @@ procedure Discriminated_Record is
 begin
    Inst1.A (5) := 1;
    Inst1.B (10) := 2;
+
+   pragma Assert (Inst1.A (5) = 1);
```
martin-cs: This I very much approve of.  If you update `B` and then check the same assertion again it would be even better!

tjj2017:
The changes look sensible and should be satisfactory when the comments left by other reviewers are addressed.

PR #292, opened 2019-09-23
Refactor/tree walk part 1

hannes-steffenhagen-diffblue:
This is part 1 of the changes to Tree_Walk. Since this is a very large file the changes come in multiple parts

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -1913,10 +1913,11 @@ package body Tree_Walk is
             return First_Alt_Test;
          end if;
          declare
-            Big_Or : constant Irep := New_Irep (I_Op_Or);
+            Big_Or : constant Irep := Make_Op_Or
+              (Source_Location => Sloc (This_Alt),
+               I_Type => CProver_Bool_T);
          begin
             Append_Op (Big_Or, First_Alt_Test);
-            Set_Type (Big_Or, New_Irep (I_Bool_Type));
             while Present (This_Alt) loop
                Append_Op (Big_Or, Make_Single_Test (This_Alt));
                Next (This_Alt);
@@ -1967,61 +1968,25 @@ package body Tree_Walk is
    function Do_Loop_Statement (N : Node_Id) return Irep is
       Iter_Scheme  : constant Node_Id := Iteration_Scheme (N);
       Body_Block   : constant Irep := Process_Statements (Statements (N));
-
-      function Do_For_Statement (Param : Irep; Cond : Irep; Post : Irep)
-                                 return Irep;
-
-      function Do_While_Statement (Cond : Irep) return Irep;
-
-      ----------------------
-      -- Do_For_Statement --
-      ----------------------
-
-      function Do_For_Statement (Param : Irep; Cond : Irep; Post : Irep)
-                                 return Irep
-      is
-         Ret : constant Irep := New_Irep (I_Code_For);
-      begin
-         Set_Source_Location (Ret, Sloc (N));
-         Set_Init (Ret, Param);
-         Set_Cond (Ret, Cond);
-         Set_Iter (Ret, Post);
-         --  body block done in caller
-         return Ret;
-      end Do_For_Statement;
-
-      ------------------------
-      -- Do_While_Statement --
-      ------------------------
-
-      function Do_While_Statement (Cond : Irep) return Irep is
-         Ret : constant Irep := New_Irep (I_Code_While);
-      begin
-         --  body block done in caller
-         Set_Source_Location (Ret, Sloc (N));
-         Set_Cond (Ret, Cond);
-         return Ret;
-      end Do_While_Statement;
-
       Loop_Irep : Irep;
-      Loop_Wrapper : constant Irep := New_Irep (I_Code_Block);
+      Loop_Wrapper : constant Irep := Make_Code_Block
+        (Sloc (N));
    begin
       if not Present (Iter_Scheme) then
-         --  infinite loop
-         declare
-            Const_True : constant Irep := New_Irep (I_Constant_Expr);
-         begin
-            Set_Value (Const_True, "true");
-            Set_Type (Const_True, Make_Bool_Type);
-            Loop_Irep := Do_While_Statement (Const_True);
-         end;
+         Loop_Irep := Make_Code_While
+           (Loop_Body => Body_Block,
+            Cond => CProver_True,
+            Source_Location => Sloc (N));
       else
          if Present (Condition (Iter_Scheme)) then
             --  WHILE loop
             declare
                Cond : constant Irep := Do_Expression (Condition (Iter_Scheme));
             begin
-               Loop_Irep := Do_While_Statement (Cond);
+               Loop_Irep := Make_Code_While
+                 (Loop_Body => Body_Block,
+                  Cond => Cond,
+                  Source_Location => Sloc (N));
             end;
          else
             --  FOR loop.
@@ -2060,7 +2025,7 @@ package body Tree_Walk is
                        I_Type          => Type_Loopvar,
                        Identifier      => Loopvar_Name);
 
-                  Init : constant Irep := New_Irep (I_Code_Assign);
+                  Init : Irep;
                   Cond : Irep;
                   Post : Irep;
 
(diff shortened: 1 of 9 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -493,12 +492,10 @@ package body Tree_Walk is
...
 
    function Do_Address_Of (N : Node_Id) return Irep is
-      R : constant Irep := New_Irep (I_Address_Of_Expr);
    begin
       if not (Kind (Get_Type (Do_Expression (Prefix (N)))) in Class_Type) then
-         Report_Unhandled_Node_Empty (N, "Do_Address_Of",
-                                      "Kind not in class type");
-         return R;
+         return Report_Unhandled_Node_Irep (N, "Do_Address_Of",
+                                            "Kind not in class type");
```
xbauch: This will fail the post-condition, not that it matters too much.

hannes-steffenhagen-diffblue: Hm, yeah. This is kind of the problem with our pre/postconditions, they're much tighter than what we can actually guarantee without breaking actual invariants

gnat2goto/driver/tree_walk.adb
```diff
@@ -695,28 +692,24 @@ package body Tree_Walk is
...
-                          Global_Symbol_Table));
-            end;
-         else
-            Set_Rhs (R, Typecast_If_Necessary (
-                     RHS, Get_Type (LHS), Global_Symbol_Table));
-         end if;
-         return R;
+         return Make_Code_Assign
+           (Lhs => LHS,
+            Rhs => RHS,
```
xbauch: I'm pretty sure the typecasts were there for a reason, but it's nice it works without them.

hannes-steffenhagen-diffblue: I didn't intentionally remove them. Ohps. Do you think I should add them back? If the presence or absence of these type casts can make a difference we should have a test to show it, and we apparently don't right now, which is a bit of a shame

gnat2goto/driver/tree_walk.adb
```diff
@@ -861,48 +861,36 @@ package body Tree_Walk is
          end;
       end Make_Case_Test;
 
-      Case_Body_Leaf : Irep := Ireps.Empty;
-      Case_Body_Root : Irep := Ireps.Empty;
-      This_Alt : Node_Id := First (Alternatives (N));
+      function Make_Case_If_Expression (Alternatives : Node_Id) return Irep;
+      function Make_Case_If_Expression (Alternatives : Node_Id) return Irep
+      is
+         This_Alt : constant Node_Id := Alternatives;
```
xbauch: Do you still need this node?

hannes-steffenhagen-diffblue: So, technically not functionality wise, but I wanted to rename it to distinguish between the list of alternatives and the thing that's currently pointing to the head of that list (even though they're the same thing as far as the implementation is concerned)

gnat2goto/driver/tree_walk.adb
```diff
@@ -1531,38 +1520,42 @@ package body Tree_Walk is
...
+                     Make_Range_Expression
+                       (Sym_Nondet,
+                        Get_Bound (N, Followed_Type, Bound_Low),
+                        Get_Bound (N, Followed_Type, Bound_High)));
+               else
+                  return Make_Assume_Expr
+                    (N,
+                     Make_Constant_Expr
+                       (Source_Location => Source_Location,
+                        I_Type          => Make_Bool_Type,
```
xbauch: `CProver_Bool_T`

gnat2goto/driver/tree_walk.adb
```diff
@@ -1377,12 +1350,12 @@ package body Tree_Walk is
    function Do_And_Then (N : Node_Id) return Irep is
       L : constant Node_Id := Left_Opnd (N);
       R : constant Node_Id := Right_Opnd (N);
-      Expr : constant Irep := New_Irep (I_Op_And);
+      Expr : constant Irep := Make_Op_And
+        (I_Type => CProver_Bool_T,
```
NlightNFotis: Also, that has nothing to do with your PR, but I do find it weird that our constructors for binary operators do not get the operands and we have to add them through `Append_Op`. At least it's surprising given that the other constructors do get their operands in their interface.

gnat2goto/driver/tree_walk.adb
```diff
@@ -1455,22 +1428,38 @@ package body Tree_Walk is
...
 
    function Make_Assume_Expr (N : Node_Id; Assumption : Irep) return Irep is
+      Source_Location : constant Source_Ptr := Sloc (N);
+      Assume_Params : constant Irep := Make_Parameter_List;
+      Assume_Type : constant Irep := Make_Code_Type
+        (Parameters => Assume_Params,
+         Ellipsis => False,
+         Return_Type => CProver_Void_T,
+         Inlined => False,
+         Knr => False);
```
NlightNFotis: On a related note, I've never seen anywhere where we set this to `True`. Could we maybe change the constructor to give this a default value?

gnat2goto/driver/tree_walk.adb
NlightNFotis: Why?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1723,9 +1709,10 @@ package body Tree_Walk is
          if Nkind (E) in N_Has_Etype and then (Is_Type (Etype (E))) then
             return Do_Defining_Identifier (E);
          end if;
-         Report_Unhandled_Node_Empty (N, "Do_Identifier",
-                                         "Etype not a type");
-         return R : constant Irep := New_Irep (I_Symbol_Expr);
```
NlightNFotis: This was previously returning a `Symbol_Expr`. Why did this change?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1763,9 +1732,11 @@ package body Tree_Walk is
...
       end if;
       if Present (Condition (N)) then
-         return R : constant Irep := Do_When (Condition (N)) do
-            Set_Then_Case (R, Jump_Irep);
-         end return;
+         return Make_Code_Ifthenelse
+           (Source_Location => Sloc (N),
+            Cond => Do_Expression (Condition (N)),
+            Then_Case => Jump_Irep,
+            Else_Case => Ireps.Empty);
```
NlightNFotis: `Make_Nil`? Or am I wrong?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1815,16 +1786,11 @@ package body Tree_Walk is
...
-      begin
-         Set_Source_Location (Ret, Sloc (N));
-         Set_Cond (Ret, Cond_Expr);
-         Set_Then_Case (Ret, If_Block);
-         return Ret;
-      end Do_If_Block;
+         (Make_Code_Ifthenelse
+           (Cond => Do_Expression (Condition (N)),
+            Then_Case => Process_Statements (Then_Statements (N)),
+            Else_Case => Ireps.Empty,
```
NlightNFotis: Same?

testsuite/gnat2goto/tests/expanded_name/expanded_name.adb
```diff
@@ -9,5 +9,5 @@ begin
       when others =>
          N := 1;
    end case;
-   pragma Assert ((if Withed_Package.X = 1 then N = 1));
+   pragma Assert ((if Withed_Package.X = 1 then N = 2 else N = 1));
```
NlightNFotis: Excellent catch.

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: TBH we could probably just nuke it entirely and only emit the default value when generating JSON, but I'm not putting that into this PR

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: I'm mirroring the previous code where they were left empty. Not sure if it'd also work with nil (the thing about nil is that it's actually _something_ that is nil, whereas with empty that just means the value isn't set to anything)

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: There isn't really a way to create a valid symbol expr without setting its name and type, but we can't do this here because this is an error case.

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: As per in-person talk: Our `And` and `Or` operators are actually n-ary operators, hence they take a list of operands rather than having them explicitly passed (I don't particularly like this either, I'm considering adding an overload to `goto-utils` that takes an array of operands in cases where we know which ones we want).

PR #293, opened 2019-09-25
Add basic support for function pointers

xbauch:
that is handling the access-function-definition node and the case when function-call node is explicit-dereference.

This required some special casing for getting the formal parameter (in gnat-utils) and building the address-of for functions.

A test is included.

### DIFF
--- a/experiments/golden-results/muen-summary.txt
+++ b/experiments/golden-results/muen-summary.txt
@@ -1,29 +1,44 @@
-Occurs: 135 times
+Occurs: 173 times
 Calling function: Process_Declaration
 Error message: Representation clause unsupported: alignment
 Nkind: N_Attribute_Definition_Clause
+--
+Occurs: 160 times
+Calling function: Process_Declaration
+Error message: Use type clause declaration
+Nkind: N_Use_Type_Clause
 --
 Occurs: 112 times
 Calling function: Process_Declaration
 Error message: Exception declaration
 Nkind: N_Exception_Declaration
 --
-Occurs: 104 times
+Occurs: 105 times
 Calling function: Do_Type_Definition
 Error message: Access type unsupported
 Nkind: N_Access_To_Object_Definition
 --
-Occurs: 64 times
+Occurs: 66 times
 Calling function: Do_Private_Type_Declaration
 Error message: Abstract type declaration unsupported
 Nkind: N_Private_Type_Declaration
 --
-Occurs: 28 times
+Occurs: 56 times
+Calling function: Process_Declaration
+Error message: size clause not applied by the front-end
+Nkind: N_Attribute_Definition_Clause
+--
+Occurs: 32 times
 Calling function: Process_Declaration
 Error message: Abstract subprogram declaration
 Nkind: N_Abstract_Subprogram_Declaration
 --
-Occurs: 12 times
+Occurs: 13 times
+Calling function: Do_Derived_Type_Definition
+Error message: record extension unsupported
+Nkind: N_Derived_Type_Definition
+--
+Occurs: 13 times
 Calling function: Process_Declaration
 Error message: Unknown declaration kind
 Nkind: N_Private_Extension_Declaration
@@ -49,11 +64,6 @@ Error message: This type of library_unit is not yet handled
 Nkind: N_Generic_Package_Declaration
 --
 Occurs: 8 times
-Calling function: Process_Declaration
-Error message: Use type clause declaration
-Nkind: N_Use_Type_Clause
---
-Occurs: 8 times
 Calling function: Process_Pragma_Declaration
 Error message: Unknown pragma
 Nkind: N_Pragma
@@ -74,14 +84,9 @@ Error message: Unsupported pragma: Ada 05
 Nkind: N_Pragma
 --
 Occurs: 6 times
-Calling function: Do_Address_Of
-Error message: Kind not in class type
-Nkind: N_Attribute_Reference
---
-Occurs: 6 times
-Calling function: Do_Identifier
-Error message: Etype not a type
-Nkind: N_Identifier
+Calling function: Do_Type_Reference
+Error message: Type of type not a type
+Nkind: N_Defining_Identifier
 --
 Occurs: 6 times
 Calling function: Process_Declaration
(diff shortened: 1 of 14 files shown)


experiments/golden-results/muen-summary.txt
```diff
@@ -3562,8 +3557,13 @@ Occurs: 1 times
...
+--
+Occurs: 4 times
+<========================>
+raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed postcondition from tree_walk.ads:84
+
 --
 Occurs: 2 times
 <========================>
-raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed precondition from ireps.ads:1597
+raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed precondition from arrays.ads:57
```
hannes-steffenhagen-diffblue: Any idea why the assertion failures were swapped around and changed?

gnat2goto/driver/gnat_utils.adb
```diff
@@ -11,6 +11,9 @@ package body GNAT_Utils is
       Nam : constant Node_Id := Name (N);
 
    begin
+      if Nkind (Nam) = N_Explicit_Dereference then
```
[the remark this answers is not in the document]
xbauch: Because the case expression always returns an `Entity` of something. In the case of `explicit-dereference` the entity-of-prefix-of-name-of-n leads to the return type of the function (while we need the whole type of the function).

gnat2goto/driver/tree_walk.adb
```diff
@@ -4475,8 +4475,17 @@ package body Tree_Walk is
...
 
    function Do_Type_Reference (E : Entity_Id) return Irep is
+      Type_Name : constant String := Unique_Name (E);
+      --  will become unique-name of:
+      --  (if Ekind (E) = E_Access_Subtype then Etype (E) else E);
+      --  in the next commit
+      Type_Id : constant Symbol_Id := Intern (Type_Name);
    begin
-      return Make_Symbol_Type (Identifier => Unique_Name (E));
+      if Global_Symbol_Table.Contains (Type_Id) then
```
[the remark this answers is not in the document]
xbauch: I would like this to always be an error.

gnat2goto/driver/gnat_utils.adb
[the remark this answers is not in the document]
hannes-steffenhagen-diffblue: It might be cleaner to move this into the case expression, and move the `Entity` call from outside the case to the cases that need it

gnat2goto/driver/tree_walk.adb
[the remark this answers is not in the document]
xbauch: Right now, only the `primitive-pointer` test fails here (seems that when we're doing the object-declaration we assume that the type of the object was already defined; but we don't need to explicitly define access-types).

experiments/golden-results/muen-summary.txt
xbauch: Maybe: before, the `do-type-reference` always return `i-symbol-type`. Now it may return a `symtype` of a symbol, regardless of what kind of irep that `symtype` is. It could, for example, be the case that when buiilding the symbol we ran into an unsupported feature and returned a nil-irep, which was then stored as a `symtype`.

gnat2goto/driver/gnat_utils.adb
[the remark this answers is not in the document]
xbauch: Done.

xbauch:
@martin-cs There aren't many instances of function pointers in our golden results. Hannes suggested that Ada function pointers get more difficult when using nested functions, etc. Do you think this solution is sufficient for now?

experiments/golden-results/muen-summary.txt
```diff
@@ -3562,8 +3557,13 @@ Occurs: 1 times
 +===========================GNAT BUG DETECTED==============================+
 | GNU Ada (ada2goto) Constraint_Error Symbol_Table_Info.Symbol_Maps.Constant_Reference: key not in map|
 Error detected at REDACTED
+--
+Occurs: 4 times
+<========================>
+raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed postcondition from tree_walk.ads:84
+
```
martin-cs: Higher up you removed 3 instances of Do_Type_Definition / Unknown expression kind / N_Access_Function_Definition .  Here there are 4 extra assertion failures.  Forgive me but it doesn't look like this is a net improvement.  Also, where is the extra error coming from?

martin-cs:
@xbauch : I am fine with minimal solutions as long as the more complex cases that are not handled result in an unsupported feature being raised.  It is not clear to me where in that patch this would happen.  Note that the compiler's type checking and visibility rules will have already located / blocked some of the difficult cases.

xbauch:
@martin-cs I've added support for procedures and turned some assertion into unsupported-features, so that's it's easier to see what caused them. As per the _more complex cases_ we should be explicitly reporting: do you have something specific in mind?

gnat2goto/driver/tree_walk.adb
```diff
@@ -1289,8 +1308,28 @@ package body Tree_Walk is
...
                                                      "Range attribute");
-               when Attribute_First  => return Do_Array_First (N);
-               when Attribute_Last   => return Do_Array_Last (N);
+               when Attribute_First  =>
+                  if Nkind (Etype (Prefix (N))) = N_Defining_Identifier
+                    and then
+                      Get_Name_String (Chars (Etype (Etype (Prefix (N)))))
+                    = "string"
+                  then
+                     return Report_Unhandled_Node_Irep (N, "Do_Expression",
```
martin-cs: These errors seem to be triggered in UKNI which, I don't think, includes function access types.  Are they solving a different problem, if so, given they are being triggered, are they solving it correctly?

martin-cs:
@xbauch : more complex cases, some are structural, for example, when the  variable is a member of a record, an element of an array, a combination of the above.  I think this code should handle these?  The other area of complexity is the interaction of access types and subtyping.  For example if AS is a subtype of A and BS is a subtype of B, can you assign a function with signature B -> AS to a access type of signature BS -> A?  I don't think you need to worry about this right now (maybe @tjj2017 has some thoughts on this) as long as you know where it would be caught.  It may be that the GNAT front-end does enough.

gnat2goto/driver/tree_walk.adb
xbauch: These error were always there, just hidden as raised exceptions. They are unrelated to function pointers, but our supporting of function pointers uncovered them, as exceptions. I turned some exceptions into unsupported features so that's it's easier to investigate them.

martin-cs: Brilliant; thank you.

PR #295, opened 2019-09-27
Basic exception support

xbauch:
Turning: `raise exception` into `assert(false); assume(false);` (wrapped in a nicely named function) and conditioning the handler statements by `false` , i.e.
```
begin
  Foo (5);
exception
  when My_Favourite_Exception => Destroy_Universe (Now);
end;
```
is translated into
```
Foo(5);
if (0) { Destroy_Universe(Now); }
```
There is a test (3rd commit) and the first commit is from a prerequisite #284 -- no need to review here.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -83,11 +83,6 @@ Calling function: Do_Expression
 Error message: Index of string unsupported
 Nkind: N_Indexed_Component
 --
-Occurs: 18 times
-Calling function: Process_Declaration
-Error message: Exception declaration
-Nkind: N_Exception_Declaration
---
 Occurs: 14 times
 Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: Global
@@ -102,11 +97,6 @@ Occurs: 11 times
 Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: Check
 Nkind: N_Pragma
---
-Occurs: 10 times
-Calling function: Process_Statement
-Error message: Raise statement
-Nkind: N_Raise_Statement
 --
 Occurs: 9 times
 Calling function: Process_Pragma_Declaration
@@ -1825,17 +1815,17 @@ Error detected at REDACTED
 --
 Occurs: 9 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure failed precondition from goto_utils.ads:142|
+| GNU Ada (ada2goto) Assert_Failure failed precondition from goto_utils.ads:147|
 Error detected at REDACTED
 --
 Occurs: 3 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure failed precondition from ireps.ads:1749|
+| GNU Ada (ada2goto) Assert_Failure failed precondition from ireps.ads:1750|
 Error detected at REDACTED
 --
 Occurs: 3 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure failed precondition from ireps.ads:1749|
+| GNU Ada (ada2goto) Assert_Failure failed precondition from ireps.ads:1750|
 Error detected at REDACTED
 --
 Occurs: 3 times
@@ -1845,7 +1835,7 @@ Error detected at REDACTED
 --
 Occurs: 3 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:976                      |
+| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:982                      |
 Error detected at REDACTED
 --
 Occurs: 2 times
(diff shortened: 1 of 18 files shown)


experiments/golden-results/StratoX-summary.txt
```diff
@@ -154,6 +144,11 @@ Error message: function entity not defining identifier
 Nkind: N_Function_Call
 --
 Occurs: 2 times
+Calling function: New_Function_Symbol_Entry
+Error message: __CPROVER_Ada_Raise_Exception
```
xbauch: This is unfortunate.

xbauch: Sorted.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4282,6 +4371,29 @@ package body Tree_Walk is
          Process_Statement (HSS, Reps);
       end if;
 
+      if Present (Exception_Handlers (HSS)) then
+         declare
+            A_Handler : Node_Id := First (Exception_Handlers (HSS));
```
NlightNFotis: What is this `Exception_Handlers`? I couldn't find anything relevant. Is this supplied from the Gnat frontend somehow?

gnat2goto/driver/goto_utils.adb
```diff
@@ -118,6 +119,19 @@ package body GOTO_Utils is
       return True_V;
    end CProver_True;
 
+   False_V : Irep := Ireps.Empty;
+   function CProver_False return Irep
```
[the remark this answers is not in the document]
hannes-steffenhagen-diffblue: I needed `CProver_True` but didn't need `CProver_False`. I agree it's somewhat weird, but it really comes down to whether you prefer symmetry over not adding unused code.

gnat2goto/driver/tree_walk.adb
hannes-steffenhagen-diffblue: Yeah, this appears to be a gnat-node getter.

gnat2goto/driver/goto_utils.adb
```diff
@@ -139,6 +153,32 @@ package body GOTO_Utils is
          (I_Subtype => Base,
           Width => Pointer_Type_Width));
 
+   function Make_Array_String_Type (Size : Integer) return Irep is
```
hannes-steffenhagen-diffblue: Maybe `Make_Constant_Length_String_Type` or something, because this is specifically for known-length strings

xbauch:
@martin-cs What do you think?

gnat2goto/driver/tree_walk.adb
```diff
@@ -3599,6 +3605,12 @@ package body Tree_Walk is
       end if;
    end Do_Package_Specification;
 
+   procedure Do_Exception_Declaration (N : Node_Id) is
+   begin
+      null;
+      --  Ignored for now
```
martin-cs: I have given this some thought and I think it is OK.

gnat2goto/driver/tree_walk.adb
```diff
@@ -4209,6 +4221,82 @@ package body Tree_Walk is
...
+      Source_Loc : constant Source_Ptr := Sloc (N);
+
+      Func_Params : constant Irep := Make_Parameter_List;
+      Func_Type : constant Irep :=
+           Make_Code_Type (Parameters  => Func_Params,
+                           Ellipsis    => False,
+                           Return_Type => CProver_Void_T,
+                           Inlined     => False,
+                           Knr         => False);
+      Function_Name : constant String := "__CPROVER_Ada_Raise_Exception";
```
martin-cs: Yes.

testsuite/gnat2goto/tests/basic_exception/test.out
```diff
@@ -0,0 +1,49 @@
+Error from cbmc test:
+warning: ignoring typecast
```
martin-cs: That's not good... why is this happening?

testsuite/gnat2goto/tests/basic_exception/test.adb
```diff
@@ -0,0 +1,26 @@
...
+      when Too_Small =>
+         Error_Count := Error_Count + 1;
+   end;
+   begin
+      More_Than_Five (5);
+   exception
+      when Too_Small =>
+         Error_Count := Error_Count + 1;
+   end;
+   --  this will be unreachable due to implementation of `raise`
```
martin-cs: Great!  That's one of the test I wanted.  Thank you.

testsuite/gnat2goto/tests/basic_exception/test.adb
```diff
@@ -0,0 +1,26 @@
+procedure Test is
+   Too_Small : exception;
```
martin-cs: Please can we also check for exceptions raised by run time checks?  Obviously these will be regarded as dead code as we don't throw the exceptions needed (or catch them...) but it would be good to know it works.

hannes-steffenhagen-diffblue: @martin-cs runtime checks are under control of the backend (i.e. us). We currently don't emit exceptions for those at all, we have specific `*_Check` handlers for those (mostly range_check).

It _might_ be worth rethinking those now that we're tackling (at least a bit of) exception handling though.

testsuite/gnat2goto/tests/basic_exception/test.out
xbauch: I'm passing the `comment` as a `char[]` but CBMC is complaining that it's incompatible with `char*`. So I typecast it to `char*` and now it's complaining that the typecast is ignored.

testsuite/gnat2goto/tests/basic_exception/test.adb
xbauch: Should it be part of this PR?

martin-cs: My apologies; I wasn't clear.  I like the way you are handling exceptions.  The function call at throw / raise and `if (0) { ... }` around the handler is good.

What I would like is one test case which tries to catch a run-time exception, say from an array bounds or the like.  This will, as @hannes-steffenhagen-diffblue observes, result in a call to the `__CPROVER_Ada_range_check` function at the array dereference and this patch will mean that the handler for the range check will be marked as dead code.  That's fine and expected.  I would just like that this happens documented in a test case.  This doesn't require you to change the handling of exceptions in this patch.

testsuite/gnat2goto/tests/basic_exception/test.out
xbauch: I couldn't reproduce this with C code alone. Maybe there is some processing when parsing C directly, that hides this cast.

xbauch:
@martin-cs I've added the test for raising range-check exception (which is handle internally).

testsuite/gnat2goto/tests/basic_exception/test.out
martin-cs: We used to have some issues with typecasts being not ignored as they should.  I think it might be worth digging a little deeper on this because errors like this used to mean the back-end was just ignoring that expression all together.

testsuite/gnat2goto/tests/basic_exception/test.adb
martin-cs: The added one is great; thank you.

martin-cs:
@xbauch Looks good apart from that issue with typecasts being ignored.  If we can definitely say that that is harmless then fine by me.

xbauch:
> @xbauch Looks good apart from that issue with typecasts being ignored. If we can definitely say that that is harmless then fine by me.

@martin-cs Very creepy fix; I wrote a similar C code and in goto the string constant was passed as an address-of-index-expr (the index being zero). So I did the same, and now the solver flattening no longer complains.

gnat2goto/driver/goto_utils.adb
```diff
@@ -136,14 +135,14 @@ package body GOTO_Utils is
    end Int64_T;
 
    Char_T : Irep := Ireps.Empty;
-   function Uint8_T return Irep
+   function Int8_T return Irep
```
hannes-steffenhagen-diffblue: This `uint8_t` doesn't have a replacement because this was the only use for it (as per correspondence with @xbauch )

gnat2goto/driver/goto_utils.adb
```diff
@@ -225,12 +224,14 @@ package body GOTO_Utils is
    function Make_Array_String_Type (Size : Integer) return Irep is
    begin
       return Make_Array_Type
-        (I_Subtype => Uint8_T,
+        (I_Subtype => Int8_T,
          Size      => Make_Constant_Expr
            (Source_Location => Internal_Source_Location,
-            I_Type          => Int32_T,
+            I_Type          => Uint32_T,
```
hannes-steffenhagen-diffblue: Why not `CProver_Size_T`?

gnat2goto/ireps/irep_specs/index_expr.json
```diff
@@ -1,5 +1,6 @@
 {
   "parent": "expr",
+  "id": "array",
```
hannes-steffenhagen-diffblue: Why not `id: "index"`?

PR #299, opened 2019-10-03
Refactor/make new irep private

hannes-steffenhagen-diffblue:
From now on code outside of the internals of the Ireps module will no longer compile if it tries to use `New_Irep`.

Therefore it is now impossible to create an irep without all of its fields being initialised (they may still be initialised to dummy values, and this doesn't touch the setters, so this doesn't provide full type safety for ireps but it does at least make it a *lot* harder to create invalid ireps by accident, or introduce changes to the irep structure that break code without us noticing.

### DIFF
--- a/gnat2goto/ireps/ireps_generator.py
+++ b/gnat2goto/ireps/ireps_generator.py
@@ -1581,8 +1581,8 @@ class IrepsGenerator(object):
         write(b, "")
 
 
-        write(s, "function New_Irep (Kind : Valid_Irep_Kind) return Irep;")
-        write(s, "")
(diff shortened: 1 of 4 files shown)


gnat2goto/ireps/ireps_generator.py
```diff
@@ -1581,8 +1581,8 @@ def generate_code(self, optimize, schema_file_names):
         write(b, "")
 
 
-        write(s, "function New_Irep (Kind : Valid_Irep_Kind) return Irep;")
```
NlightNFotis: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAaa

PR #301, opened 2019-10-07
Add property class and function name to assertions

hannes-steffenhagen-diffblue:
These are used by cbmc to pretty print assertions. As a result, nearly all tests
have slightly different outputs now. This doesn't have the actual pretty printed
assertion expressions yet, but these are planned

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -2200,6 +2200,42 @@ package body Tree_Walk is
       is
          Which : constant Pragma_Id := Get_Pragma_Id (N_Orig);
 
+         function Make_Assert_Comment (Condition : Irep) return Irep;
+
+         function Make_Assert_Comment (Condition : Irep) return Irep
+         is
+            --  will be used to pretty print condition later
+            pragma Unreferenced (Condition);
+
+            --  return name of containing function or package
+            --  (whichever comes earlier)
+            function Get_Context_Name (Intermediate_Node : Node_Id)
+                                      return String;
+            function Get_Context_Name (Intermediate_Node : Node_Id)
+                                      return String is
+            begin
+               case Nkind (Intermediate_Node) is
+                  when N_Subprogram_Body | N_Package_Body =>
+                     return Get_Name_String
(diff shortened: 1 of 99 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -2200,6 +2200,42 @@ package body Tree_Walk is
...
+               end case;
+            end Get_Context_Name;
+
+            Source_Loc : constant Irep := Get_Source_Location (N);
+            Context_Name : constant String := Get_Context_Name (N);
+            Comment_Prefix : constant String := "assertion ";
+         begin
+            Set_Property_Class (Source_Loc, "assertion");
+            Set_Function (Source_Loc, Context_Name);
+            Set_Comment (Source_Loc, Comment_Prefix);
```
hannes-steffenhagen-diffblue: ^- yes, this does go a bit against the previous PR which mostly removed use cases for setters, but I couldn't see a good way around it in this case

hannes-steffenhagen-diffblue:
Also, the source files disappearing from the assertion messages is a CBMC thing; The info is still there, cbmc just chooses not to display the file name when a function name is present apparently

testsuite/gnat2goto/tests/absolute/test.out
```diff
@@ -1,4 +1,4 @@
-[1] file absolute.adb line 7 assertion: SUCCESS
-[2] file absolute.adb line 8 assertion: SUCCESS
-[3] file absolute.adb line 10 assertion: FAILURE
+[absolute.assertion.1] line 7 assertion : SUCCESS
```
xbauch: Why the space after "assertion"?

hannes-steffenhagen-diffblue: Good question, it doesn't make a lot of sense here but the final format is supposed to be `assertion {pretty printed condition}:`, so I included it here. I can remove it and readd it in the next commit if it bothers you though.

xbauch: Or that's where the actual assertion will live? Ok.

PR #302, opened 2019-10-08
Feature/pretty assertions part 2

hannes-steffenhagen-diffblue:
This currently reads the actual file and does some miniscule amount of parsing
to extract the actual source code. I had considered trying a pretty printer, but
the one in gnat right now seems to be quite limited and hard to use, and getting
the output exactly the way we want would be a bit hard using that, and writing
another pretty printer from scratch would be a pretty monumental task. This
approach looks a bit hacky, but should work in all cases thanks to Ada's syntax
being very regular.

Right now there's a bit of a performance issue in that we actually have to scan
through the file several times every time we see a '-' character; This can be
avoided by either using a different File API or by restructuring the code if
needed (I'd consider this version more of a PoC, although in practice I expect
the performance impact will be tolerable even on fairly large projects).

**The first commit is part of a separate PR, do not review here**

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -1705,7 +1705,7 @@ Raw compiler error message:
 --
 Occurs: 29 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:4905                     |
+| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:5051                     |
 Error detected at REDACTED
 --
 Occurs: 6 times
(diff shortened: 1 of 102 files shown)


testsuite/gnat2goto/tests/arrays_access/test.c
```diff
@@ -0,0 +1,15 @@
+#include <assert.h>
```
xbauch: Yes, but why?

hannes-steffenhagen-diffblue: Accidental include. Will remove :+1:

hannes-steffenhagen-diffblue:
@chrisr-diffblue yes, ifs and such things wouldn't be a problem for this, for example this:

```
  pragma Assert (3 +
    (if 5 < 2
     then 3 *
        (if 2 < 1
         then 9
         else 8)
    else 42-23)
    < 19);
```

turns into this:

```
[test.assertion.1] line 3 assertion 3 + (if 5 < 2 then 3 * (if 2 < 1 then 9 else 8) else 42-23) < 19: FAILURE
```

Compare with C/CBMC:

```
  assert (3 +
      (5 < 2
       ? 3 *
       (2 < 1
        ? 9
        : 8)
       : 42-23)
      < 19);
```

```
[main.assertion.1] line 3 assertion 3 + (5 < 2 ? 3 * (2 < 1 ? 9 : 8) : 42-23) < 19: FAILURE
```

martin-cs:
@hannes-steffenhagen-diffblue : I do not mean to disrespect your opinion but ... was there really no way that you couldn't do this with pretty printing?  I am concerned that with a more complex build system / series of inclusions this could go wrong.

PR #303, opened 2019-10-09
Breakup building script list_unsupported.sh

NlightNFotis:


experiments/check_env_sanity.cfg
```diff
@@ -0,0 +1,88 @@
+#!/bin/sh
```
hannes-steffenhagen-diffblue: This file shouldn't be called "*.cfg". This is a commonly used extension for config files, having a script with that name is going to confuse people.

experiments/list_unsupported.sh
```diff
@@ -16,89 +16,10 @@ usage()
...
-gplusplus=`command -v g++`
-if ! ${gplusplus} --std=c++14 "${experiment_dir}/collect_unsupported.cpp" -o CollectUnsupported ; then
-   echo >&2 "Failed to compile support tool 'CollectUnsupported' using ${gplusplus}"
-   echo >&2 "You need a version of g++ on the PATH that supports C++14"
-   exit 8
-fi
-export PATH="${saved_path}"
-
-echo >&2 "...environment is OK."
+running_dir=$(dirname $(readlink -f $0))
```
hannes-steffenhagen-diffblue: Should be surrounded with quotes (both commands), also @chrisr-diffblue correct me if I'm wrong but I believe `readlink -f` doesn't work on OSX?

experiments/list_unsupported.sh
```diff
@@ -16,89 +16,10 @@ usage()
...
 
+# include environment sanity checking module and basic config vars
```
hannes-steffenhagen-diffblue: This should mention _which_ vars this is setting at least

experiments/list_unsupported.sh
chrisr-diffblue: Correct - the `-f` means something different on OSX (and the `readlink`  tool is not in POSIX either). However, given its use here, I'm not quite sure why the readlink is even needed? I'd just use something like `. "$(dirname $0)/check_env_sanity.cfg"`

chrisr-diffblue:
Or perhaps the work flow would look something like:
`sanity_check_env.sh && "build command" > build.stdout 2>build.stderr ; list_unsupported.sh build.stdout build.stderr`

experiments/list_unsupported.sh
hannes-steffenhagen-diffblue: @chrisr-diffblue the `readlink` is necessary to get the correct path if the tool is invoked via symlink. In this case this precaution is probably not necessary as it is unlikely to be exported to a global location.

experiments/check_env_sanity.cfg
NlightNFotis: Agreed. It started life as only having some variables, but it outgrew this use case, without an appropriate rename following it.

martin-cs:
@hannes-steffenhagen-diffblue : as @chrisr-diffblue says, `list_unsupported.sh` has two bits of functionality, a psuedo-build system and collating and summarising build failures and unsupported features.  We want to be able to use both together but also to use an applications' build system and our current result summarisation.  For this the functionality needs to be split into two different interfaces.

experiments/build_vars.cfg
```diff
@@ -0,0 +1,17 @@
...
+GNAT2GOTO=`command -v gnat2goto`
+ADA_HOME=`command -v gnat`
+ADA_HOME=$(cd "$(dirname ${ADA_HOME})/.." 2>/dev/null && pwd)
+PLATFORM=`${ADA_HOME}/bin/gcc -dumpmachine`
+DEF_ADA_INCLUDE_PATH="${ADA_HOME}/lib/gcc/${PLATFORM}"
+ADA_GCC_VERSION=`${ADA_HOME}/bin/gcc -dumpversion`
+DEF_ADA_INCLUDE_PATH="${DEF_ADA_INCLUDE_PATH}/${ADA_GCC_VERSION}/rts-native/adainclude:${ADA_HOME}/include"
+export ADA_INCLUDE_PATH="${ADA_INCLUDE_PATH:-$DEF_ADA_INCLUDE_PATH}"
+
+file_name=$(basename "${project_dir}")
```
chrisr-diffblue: I think `file_name` should go away - it's an artefact of our current list_unsupported.sh script, but we wouldn't expect the users build system to know it needs to generate this special file name...

experiments/list_unsupported.sh
```diff
@@ -145,213 +67,12 @@ if [ -z "$project_dir" ]; then
...
-total_count=`grep -c -E '^---------- COMPILING: ' "$file_name".txt`
-fail_count=`grep -c -E '^---------- FAILED ' "$file_name".txt`
-non_compile_error_fail_count=`grep -c -E '^gnat2goto exit code: [0-9]*' "$file_name".txt`
-missing_feature_count=`grep -c -E '^---------- MISSING FEATURES ' "$file_name".txt`
-ok_count=`grep -c -E '^---------- OK ' "$file_name".txt`
+# Invoke the building script on project
+. "$(dirname $0)/list_unsupported_build.sh"
 
+# Invoke the report generating script on project
+. "$(dirname $0)/list_unsupported_report.sh" "$file_name".stdout "$file_name".stdout
```
chrisr-diffblue: This still makes the implicit assumption that the `list_unsupported_build.sh` knows it needs to generate `"$file_name".stdout`, etc. I'd much rather just see:
`. "$(dirname $0)/list_unsupported_build.sh" > "blah.stdout" 2> "blah.stderr"`

experiments/list_unsupported_build.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/sh
+
+echo >&2 "Project to build: ${project_dir}"
```
chrisr-diffblue: We probably want the `${project_dir}` passed as a command line argument, rather than as a magic global environment variable.

experiments/list_unsupported_build.sh
```diff
@@ -0,0 +1,76 @@
...
+
+for foldername in $(find ${path} -type d -name "*" | LC_ALL=posix sort ); do
+   count=`ls -1 ${foldername}/*.${spec_ext} 2>/dev/null | wc -l`
+   if [ $count != 0 ]
+   then
+      include_path="${include_path} -I ${foldername}"
+   fi
+done
+
+echo "${project_dir}: Unsupported features\n" > "$project_name".txt
```
chrisr-diffblue: I think that  _all_ the places in this script where it currently does redirection to a file of some type should just go to stdout/stderr as normal. Again, the end user will not want/be able to alter their build system to send stuff to various separate files so this "fake" build script should do the same - i.e. all output is on either stderr or stdout, and never to a magic file.

experiments/list_unsupported_report.sh
```diff
@@ -0,0 +1,144 @@
...
+# Include configuration variables.
+. "$(dirname $0)/build_vars.cfg"
+
+# The script reads two input files, one for standard output and one
+# for standard input. Then concatenates them into the same file. This
+# is done to avoid interleaved text between the two streams breaking
+# the parsing done.
+STDOUT=$1
+STDERR=$2
+file_name="$(cat $STDOUT $STDERR)"
```
chrisr-diffblue: This looks very bogus.... :-) This is reading the _content_ of the stdout and stderr files into a single shell variable, called `file_name` - however in the next couple of lines you then _use_ that shell variable as if it contains the name of a file.....

experiments/list_unsupported_report.sh
```diff
@@ -0,0 +1,144 @@
...
+quoted_body_ext=$(printf "%s" "${body_ext}" | sed 's/\./\\./g')
+# This redacting system is really pretty crude...
+sed '/^\[/ d' < "$file_name".txt | \
+   sed 's/"[^"][^"]*"/"REDACTED"/g' | \
+      sed "s/[^ ][^ ]*\.${quoted_body_ext}/REDACTED.${body_ext}/g" | \
+         sed "s/[^ ][^ ]*\.${quoted_spec_ext}/REDACTED.${spec_ext}/g" \
+   > "$file_name"_redacted.txt
+
+# Collate and summarise unsupported features
+LC_ALL=posix ./CollectUnsupported "$file_name".txt
```
chrisr-diffblue: For the call to `CollectUnsupported` you might be best either modifying the C++ program to accept input from stdin, or cat the two stdout/stderr files into a single temporary file that you pass to the program.

NlightNFotis:
@hannes-steffenhagen-diffblue @martin-cs Can we get a new round of reviews on this?

experiments/golden-results/StratoX-summary.txt
```diff
@@ -2205,6 +2205,11 @@ Error detected at REDACTED
 --
 Occurs: 1 times
 +===========================GNAT BUG DETECTED==============================+
+| GNU Ada (ada2goto) Assert_Failure tree_walk.adb:5051                     |
```
hannes-steffenhagen-diffblue: Can the "Update golden results" commit please include information for why the golden results here have changed? I'd have expected these to not change at all.

experiments/list_unsupported_report.sh
```diff
@@ -134,7 +134,9 @@ sed -n 's/^.*:[0-9]*:[0-9]*: error: //p' "$raw_input_file" | \
          awk '/^ *[0-9]+ .*$/ { \
             count=$1; \
             raw=$0; sub(/^ *[0-9]+ /, "", raw); \
-            redacted=raw; gsub(/"[^"]+"/, "\"REDACTED\"", redacted); \
+            redacted=raw; \
+            gsub(/"[^"]+"/, "\"REDACTED\"", redacted); \
+            gsub(/at ([a-zA-Z0-9\.\-_]+:[0-9]+)/, "at REDACTED", redacted); \
```
hannes-steffenhagen-diffblue: Is this not affecting the golden results? If it is, why is there no change to the golden results _after_ this?

experiments/golden-results/StratoX-summary.txt
chrisr-diffblue: I think these were previously *missed* by the old version of the script. There seems to be a couple of cases where gnat2goto fails with an assertion failure in the middle of outputing the GNAT BUG DETECTED report - meaning it never gets to print the ending deliminator. The new script handles this better so it actually spots these cases and reports them.

experiments/list_unsupported_report.sh
chrisr-diffblue: I think we don't actually have any of the particular messages that were previously missed appearing in any of the CI projects. This was tested on the example output given by the original bug reporter.

PR #305, opened 2019-11-08
Null-statement ignored

xbauch:
Since this is a null-statement as a declaration, I'm not sure how to reproduce it.


xbauch:
@martin-cs and @hannes-steffenhagen-diffblue I've added a test and improved the commit message.

PR #306, opened 2019-11-08
Handle no return pragma

xbauch:
We insert an explicit check at the end of the function (marked by the pragma). The check is just a wrapper for `assert(false)`. Includes a check.

### DIFF
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -404,6 +404,8 @@ package body Tree_Walk is
      with Pre => Nkind (N) in N_Access_Function_Definition |
      N_Access_Procedure_Definition;
 
+   function Get_No_Return_Check return Irep;
+
    function Make_Malloc_Function_Call_Expr (Num_Elem : Irep;
                                             Element_Type_Size : Uint;
                                             Source_Loc : Irep)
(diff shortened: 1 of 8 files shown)


gnat2goto/driver/tree_walk.adb
```diff
@@ -4796,6 +4796,30 @@ package body Tree_Walk is
...
+            No_Return_Check_Args : constant Irep := Make_Argument_List;
+            No_Return_Check_Call : constant Irep := Make_Code_Function_Call
+              (Arguments       => No_Return_Check_Args,
+               I_Function      => No_Return_Check_Symbol,
+               Lhs             => Make_Nil (Internal_Source_Location),
+               Source_Location => Internal_Source_Location,
+               I_Type          => Make_Void_Type);
+         begin
+            Append_Argument (No_Return_Check_Args, Get_Int32_T_Zero);
+            Append_Op (Reps, No_Return_Check_Call);
```
hannes-steffenhagen-diffblue: This seems to add an assert at the end of a block - how does this handle returns inside of a block, like e.g.

```
procedure Test is
  procedure Does_Not_Return (X : Integer) is
  begin
    loop
      if X < 10 then
        return;
      end if;
    end loop;
  end Does_Not_Return;
  pragma No_Return (Does_Not_Return);
begin
  Does_Not_Return (5);
end Test;
```
?

xbauch: That's unfortunate.

martin-cs: Yes; this is what I was getting at when I said : 

_One thing we should be careful of; I don't know if GNAT does any checking of the pragma at a syntactic level. If not then one could attach this to a function or procedure which contains return statements and so need to check that all of these are unreachable._

My apologies for not being clearer.

xbauch:
@hannes-steffenhagen-diffblue and @martin-cs Done (the case when no-return function has a return statement).

PR #309, opened 2019-11-21
Fixes and cleanup for redaction in the list_unsupported scripts

chrisr-diffblue:


martin-cs:
@hannes-steffenhagen-diffblue They generate additional output files which are A. not used and B. harder to redact as well as the name would have you believe.

chrisr-diffblue:
Rebased now #310 has been merged.

PR #312, opened 2019-12-06
Add support for generics

hannes-steffenhagen-diffblue:
This mostly involves

A) Ignoring generic declarations
B) Ignoring generic definitions
C) Ignoring generic instantiations

Instantiations of generics are actually inserted into the AST by the frontend so
they don’t require any special handling.

This incidentally also adds rudimentary support for nested packages – they
mostly works in terms of types, procedures and functions etc, but this commit
doesn’t add initialisers for default values.


gnat2goto/driver/tree_walk.adb
```diff
@@ -5466,9 +5469,15 @@ package body Tree_Walk is
             Do_Subprogram_Body (N);
 
          when N_Package_Body =>
-            Report_Unhandled_Node_Empty (N, "Process_Declaration",
-                                         "Package body declaration");
-
+            --  XXX: Need to add instructions for initialisers of subpackage
```
xbauch: Could we have a failing test for this?

gnat2goto/driver/tree_walk.adb
```diff
@@ -994,6 +994,11 @@ package body Tree_Walk is
             --  but I can't think of one without delving deeply, and I
             --  think such uses would be unusual (TJJ 21/11/2019)
             null;
+         when N_Generic_Subprogram_Declaration
+           | N_Generic_Package_Declaration =>
+            --  Generic subprograms and packages
+            --  don't need to specially handled
```
chrisr-diffblue: Might be nice to elaborate here and say "...because they are expanded by the gnat front-end" or something like that?

gnat2goto/driver/tree_walk.adb
```diff
@@ -5093,6 +5107,10 @@ package body Tree_Walk is
                Do_Package_Declaration (N);
             when N_Package_Body =>
                null;
+            when N_Generic_Subprogram_Declaration
+              | N_Generic_Package_Declaration =>
+               --  no special handling for generics is required
```
chrisr-diffblue: and here too?

experiments/golden-results/vct-summary.txt
```diff
@@ -13,6 +13,11 @@ Calling function: Do_Aggregate_Literal
 Error message: Unhandled aggregate kind
 Nkind: N_Aggregate
 --
+Occurs: 8 times
```
chrisr-diffblue: Any idea why we see some extra missing features here? Is this just that we are now processing some generics that we were previously not processing?

hannes-steffenhagen-diffblue: @chrisr-diffblue This patch also adds rudimentary support for processing local subpackages. These weren’t processed at all before, these in particular come from a generic math package we skipped entirely before this.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -200,11 +150,6 @@ Nkind: N_Attribute_Reference
 --
 Occurs: 1 times
 Calling function: Do_Expression
-Error message: ATTRIBUTE_MIN unsupported
```
martin-cs: Why has this attribute error vanished?

experiments/golden-results/muen-summary.txt
```diff
@@ -130,7 +105,7 @@ Nkind: N_Integer_Literal
 --
 Occurs: 1 times
 Calling function: Do_Expression
-Error message: ATTRIBUTE_SIZE unsupported
+Error message: ATTRIBUTE_MIN unsupported
```
martin-cs: Another instance of the same thing?

hannes-steffenhagen-diffblue:
@martin-cs So what I _think_ is going on is that these changes lead to processing dependent packages in such a way that we get further in the processing of some of them – which in turn leads them to sometimes crash _earlier_ in cases where we do crash because of a violated precondition or things like that.

chrisr-diffblue:
@martin-cs - we're happy to merge this now if you are happy with Hannes explanation?

martin-cs:
Yes, that seems reasonable.  Please proceed.

martin-cs:
The button was right there...

PR #313, opened 2019-12-10
Ignore Warnings pragma and print name of unknown pragmas

hannes-steffenhagen-diffblue:


martin-cs:
`Process_Pragma_Declaration` as opposed to `Do_Pragma`.

martin-cs:
Also see the original ticket for an update on this task.

hannes-steffenhagen-diffblue:
FWIW these seem to be all SPARK pragmas. I’m not sure if we should ignore those, I imagine some of them could be useful?

martin-cs:
@hannes-steffenhagen-diffblue : yes, I would like to support as many of them as we can but at the moment some are outside of the kind of analysis that we do.  Note for this issue we just need the "have found unknown pragma" messages improved with what they are, we aren't necessarily handling / ignoring them.

hannes-steffenhagen-diffblue:
@martin-cs this should be all of them

PR #315, opened 2019-12-11
Feature/report unhandled aggregate kind

hannes-steffenhagen-diffblue:


martin-cs:
Please can you rebase?  It looks like you still have the null pointer PR as part of this.

hannes-steffenhagen-diffblue:
@martin-cs So in this case the problem doesn’t appear to be `Exception_Occurrence` specifically but limited private record types in general. I’ll add a test case for that.

martin-cs:
@hannes-steffenhagen-diffblue Thank you, that would help.

PR #316, opened 2019-12-12
Do size

tjj2017:
Handle S'Size and  X'Size Attributes.


gnat2goto/driver/tree_walk.adb
```diff
@@ -5840,7 +5904,7 @@ package body Tree_Walk is
                                          "Unsupported pragma: Initializes");
          when Name_Annotate |
             --  Ignore here. Rather look for those when we process a node.
-              Name_Assertion_Policy |
+              Name_Assertion_Policy | Name_Check_Policy |
```
martin-cs: Is this related?

tjj2017:
The Assertion_Policy pragma was covered by gnat2goto but in fact it never occurs in the tree because the front end rewrites it to a Check_Policy pragma.  It was just found in passing.  Not really related to the handling of the size attribute.

PR #318, opened 2019-12-13
Include function info in assertions

xbauch:
It was missing in the index/range/etc. checks. Now it is be part of get-source-location.


chrisr-diffblue:
Looks reasonable. I note that now the file names are not printed at all in the output of cbmc in the regression tests - is that expected behaviour from CBMC now, or is there some other output line that we are missing? possible due to how the test script is invoking cbmc?

chrisr-diffblue:
Answering my own question, this appears to be because the testsuite.py scripts pulls out just the assertion messages, and not any of the other output CBMC produces.

PR #320, opened 2019-12-13
Asvat modelling

tjj2017:
gnat2goto processes ASVAT models specified by an ASVAT annotation or using pragma Import.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -33,11 +33,6 @@ Calling function: Process_Pragma_Declaration
 Error message: Unknown pragma: volatile_full_access
 Nkind: N_Pragma
 --
-Occurs: 56 times
-Calling function: Do_Type_Definition
-Error message: Access type unsupported
-Nkind: N_Access_To_Object_Definition
---
 Occurs: 22 times
 Calling function: Do_Expression
 Error message: First of string unsupported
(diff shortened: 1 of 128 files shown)


tjj2017:
Update test suite and golden results.

Main change is that pragma/aspect Global is now recognised in some cases.
In Stratox there appears to be an extra failure due to a an assertion failure in tree.adb but I haven't been able to track it down.  I think it is deep in some gnat system code.
There is one case where there is a pragma Import with convention Ada which I have used to indicate an ASVAT model - this is a conflict. Perhaps I should deprecate the Import scheme and just  have the ASVAT annotation scheme?

The use of pragma Import or Export to use a foreign language reports via a Report_Unhandled_Node rather than just printing a message.  I think Martin asked me to do this to get consistent messages for the time being.

The implementation of making models in goto is incomplete but once the Do_Size and Do_Address branches are merged with this branch, gnat2goto runs, without producing errors over the exploitable example when using gprbuild exploitable_asvat.gpr,  except for a message stating that X'Valid is not supported.  I think we still need to think about how to model X'Valid.

Despite these shortcomings, If the tests and review are ok, I think this branch could be merged into master for the next release so that the example program with ASVAT modelling annotations can be run through gnat2goto.

tjj2017:
There still seems to be some trouble with results from testsuite.  On my Mac I get lots of diffs even when I run from the master branch.  I also get some diffs when I run on my Debian 10 virtual machine.  There now seem to be diffs from travis but I don't think they have anything to do with the changes I have made.  I updated 1 test results as my changes stopped a message about pragma/aspect Global not being recognised

Merry Christmas everyone - see you in the new year.

tjj2017:
This PR is closed as it has been subsumed by a later series of Mrs

PR #322, opened 2019-12-17
Update the CBMC submodule to pick up PR 5206.

martin-cs:


xbauch:
@martin-cs The ordering of test outputs changed (see #322), but I'm not too happy about the new order.

martin-cs:
@xbauch Thanks for the speedy work; I was just getting to the point of looking at what changed.  I seem to have some extra crashes but I have been hacking around with a number of things so that may well be my own fault.  I'm going to close this and future discussion can be on #322

PR #323, opened 2019-12-17
Bump CBMC to develop

xbauch:
and update the test output ordering.


martin-cs:
I see what you mean about a non-ideal order.  For me the big thing is that it is a consistent order on all platforms.  If you want another PR to get them in a more natural order I would support that but I don't think it is an immediate issue in the way that the tests failing on some platforms was.

PR #324, opened 2020-01-07
Fix object declaration

NlightNFotis:


testsuite/gnat2goto/tests/object_declaration_statement/test.opt
```diff
@@ -0,0 +1 @@
+ALL XFAIL gnat2goto fails because bounds are non constant expressions (#326)
```
martin-cs: Is it possible to change the example so that we don't have to have this as an XFAIL?  Or does that miss that not create the right parse tree?

NlightNFotis: This concern should have been addressed with a new test added as part of the latest commit.

PR #325, opened 2020-01-08
Add support to access-to-object type

xbauch:
by simply creating a pointer-to type.

Note: this somehow was not triggered by our existing primitive-pointer test (which had `int*`).

### DIFF
--- a/testsuite/gnat2goto/tests/access-to-object/main.adb
+++ b/testsuite/gnat2goto/tests/access-to-object/main.adb
@@ -1,0 +1,16 @@
+procedure Main is
+   type Day_Of_Month is range 1 .. 31;
+   type Day_Of_Month_Access is access all Day_Of_Month;
+
+   -- funny case where the subtype-indication has subtype-mark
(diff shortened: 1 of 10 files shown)


testsuite/gnat2goto/tests/access-to-object/main.adb
```diff
@@ -0,0 +1,16 @@
+procedure Main is
+   type Day_Of_Month is range 1 .. 31;
+   type Day_Of_Month_Access is access all Day_Of_Month;
+
+   -- funny case where the subtype-indication has subtype-mark
+   type C_String_Ptr is access String (1 .. Positive'Last);
```
hannes-steffenhagen-diffblue: So this is defining a pointer to string type but not actually using it??

hannes-steffenhagen-diffblue: `type Restricted_Day_Of_Month_Access is access all Day_Of_Month (range 10..15);`

xbauch: Changed.

PR #328, opened 2020-01-17
Add test for N_Loop_Parameter_Specification #28

spanners:
* tests/ukni_loop_parameter_specification
Add test.

### DIFF
--- a/testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
+++ b/testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
@@ -1,0 +1,13 @@
+procedure Test is
+
+   type Unsigned_16 is mod 2 ** 16;
+   for Unsigned_16'Size use 16;
+
+   RAM_Start : constant Unsigned_16 := 16#0200#;
+   RAM_End   : constant Unsigned_16 := 16#2199#;
+
+begin
+   for I in Unsigned_16 range RAM_Start .. RAM_End loop
+      pragma Assert (I < 16#2200#);
+   end loop;
+end Test;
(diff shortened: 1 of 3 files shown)


testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
```diff
@@ -0,0 +1,16 @@
...
+
+   type Unsigned_16 is mod 2 ** 16;
+   for Unsigned_16'Size use 16;
+
+   RAM_Start : constant Unsigned_16 := 16#0200#;
+   RAM_End   : constant Unsigned_16 := 16#2199#;
+
+begin
+   for I in Unsigned_16 range RAM_Start .. RAM_End loop
+      null;
```
chrisr-diffblue: Just so that CBMC gets exercised a bit more, is it worth adding some additional assertions into the test case, and checking that CBMC gives the expected results? e.g. asserting something about I inside the loop perhaps?

testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
```diff
@@ -0,0 +1,16 @@
+procedure Test is
+
+   type Unsigned_8 is mod 2 ** 8;
```
chrisr-diffblue: Is Unsigned_8 needed?

testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
spanners: Agreed - I have added a pragma Asesrt that `I < 16#2200#`.

testsuite/gnat2goto/tests/ukni_loop_parameter_specification/test.adb
spanners: No - I have removed it.

PR #329, opened 2020-01-21
Make the compiling steps a little more explicit

thk123:
Just a couple of things that caused me to pause when getting this compiling.

### DIFF
--- a/gnat2goto/COMPILING.md
+++ b/gnat2goto/COMPILING.md
@@ -12,7 +12,7 @@ Note: you will *not* be able to build using the debian/ubuntu gnat and gprbuild,
 For **Debian**, install the following debian packages:
   - build-essential
 
-For **Mac OSX**, you will need the XCode Command Line tools, which require XCode itself and then a 
+For **Mac OSX**, you will need the XCode Command Line tools, which require XCode itself and then a
 ```
 xcode-select --install
 ```
@@ -27,7 +27,7 @@ select your platform and then 2016 (in that order). Download
   * gnatcoll-gpl-2016
 
 
-Install the GNAT GPL Ada 2016 using the `doinstall` script inside the zip. The following instructions assume you picked `/opt/gnat` as the install destination. 
+Install the GNAT GPL Ada 2016 using the `doinstall` script inside the zip. The following instructions assume you picked `/opt/gnat` as the install destination.
 
 Install GNATCOLL GPL 2016:
   - `$ export PATH=/opt/gnat/bin:${PATH}`
@@ -38,7 +38,7 @@ Install GNATCOLL GPL 2016:
 
 # Building gnat2goto
 
-Assuming that you've built the dependencies as per the above instructions, do the following from within `gnat2goto`. 
+Assuming that you've built the dependencies as per the above instructions, do the following from within `gnat2goto`.
 
   - `$ export PATH=/opt/gnat/bin:${PATH}`
   - `$ export GPR_PROJECT_PATH=/opt/gnat/lib/gnat`
@@ -48,7 +48,7 @@ This should build into install/bin. You can test if it works by calling:
 
 ` $ install/bin/gnat2goto examples/foo.adb`
 
-If it has worked, you should have two files: `foo.ali` and `foo.json_symtab`. 
+If it has worked, you should have two files: `foo.ali` and `foo.json_symtab`.
 
 # Building and running gnat2goto unit tests
 
--- a/gnat2goto/COMPILING.md
+++ b/gnat2goto/COMPILING.md
@@ -68,7 +68,7 @@ In addition to the unit tests, gnat2goto also includes end-to-end regression
 tests. These tests aim to test the full pipeline from Ada source code input
 through to final analysis by CBMC.
 
-Instructions for running the regression test suite are described in [testsuite/gnat2goto/README.md](testsuite/gnat2goto/README.md)
+Instructions for running the regression test suite are described in [testsuite/gnat2goto/README.md](../testsuite/gnat2goto/README.md)
 
 # Running and updating missing features tests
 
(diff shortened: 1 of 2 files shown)


gnat2goto/COMPILING.md
```diff
@@ -27,7 +27,7 @@ select your platform and then 2016 (in that order). Download
   * gnatcoll-gpl-2016
 
 
-The following assumes you installed GNAT GPL 2016 into /opt/gnat using the doinstall script.
+Install the GNAT GPL Ada 2016 using the `doinstall` script inside the zip. The following instructions assume you picked `/opt/gnat` as the install destination.
```
hannes-steffenhagen-diffblue: I don’t think the doinstall script is strictly necessary, but I suppose if you didn’t want to use it you could just ignore this so OK.

chrisr-diffblue: The recommended way to install it is using the doinstall script - that's following AdaCore's instructions. In CI, we install it slightly differently via a make command, but only because that's better suited to automated install - and it's not a documented way to install it. Ahem.

chrisr-diffblue:
@thk123 - I think your CI failure might just be because Martin recently (30 mins ago) merged a PR into master. If you do a rebase + force push I *think* that might fix the "golden results".

chrisr-diffblue:
Actually, I lie. Apologies - it failed because travis couldn't checkout from one of the sample project git repos:
```
fatal: unable to access 'https://git.codelabs.ch/muen.git/': Could not resolve host: git.codelabs.ch
```
So it may be a simple restart of the travis job will get you green.

thk123:
There is a couple of things I'd like to fix in the running the regression suite that I'll add to this PR as well.

thk123:
@martin-cs @chrisr-diffblue @hannes-steffenhagen-diffblue I've added a few changes to running the regression tests section if you'd care to re-review. I will also try adding a test - so perhaps best to wait for that too!

thk123:
The adding tests instructions seemed good (apart from my inability to write a valid ada program!)

gnat2goto/COMPILING.md
```diff
@@ -66,7 +68,7 @@ In addition to the unit tests, gnat2goto also includes end-to-end regression
 tests. These tests aim to test the full pipeline from Ada source code input
 through to final analysis by CBMC.
 
-Instructions for running the regression test suite are described in [testsuite/gnat2goto/README.md](testsuite/gnat2goto/README.md)
+Instructions for running the regression test suite are described in [testsuite/gnat2goto/README.md](../testsuite/gnat2goto/README.md)
```
chrisr-diffblue: Good catch :-)

gnat2goto/COMPILING.md
```diff
@@ -93,7 +95,7 @@ one file per project. Simply copy the '<project>-summary.txt' file from
 
 # Updating the CBMC submodule
 
-As described in [testsuite/gnat2goto/README.md](testsuite/gnat2goto/README.md)
+As described in [testsuite/gnat2goto/README.md](/testsuite/gnat2goto/README.md)
```
chrisr-diffblue: Should that also be `../testsuite/.....` ?

testsuite/gnat2goto/README.md
```diff
@@ -89,6 +89,24 @@ Requirements
 
 Running tests
 -------------
+
+Ensure that both gnat2goto and cbmc are on the path:
+
+```
+export PATH=${PATH}:/path/to/gnat2goto/lib/cbmc/build/bin:/path/to/gnat2goto/gnat2goto/install
```
chrisr-diffblue: That example path for CBMC is only valid if you built with cmake and specifed `build` as the build directory. If you built CBMC using make, it'd be something like `/path/to/gnat2goto/lib/cbmc/src/cbmc` I think?

testsuite/gnat2goto/README.md
```diff
@@ -89,6 +89,24 @@ Requirements
...
+From within [gnat2goto/testsuite/gnat2goto](.) run:
+
+```
+$ ./testsuite.py -j 4
+```
+where ```-j``` sets the number of tests to run in parallel (default is 1).
+
+This will output which tests pass and fail. If a test fails, you can inspect the output 
+inside `out`. Each test will produce a file called `test_name.diff` which contains a git
+diff between the expected an actual output.
```
chrisr-diffblue: You can also add other options to `testsuite.py` such as `--timeout 60 --diffs --enable-color` (taken from CI) - which means "timeout any single test case that takes longer than 60 seconds, output diffs for any XFAIL, FAIL, or UPASS result, and use colour in the output"

testsuite/gnat2goto/README.md
martin-cs: Yes it is and please can we have that version because not all upstream users use cmake.

thk123: Can do - though not all upstream users use make either (points at self). I figured the above was clear that it refers to wher ever the build is (whereas pointing at `src` you might wrongly think it needs to be the src tree). 

Will amend to:

> ```
> export PATH=${PATH}:/path/to/gnat2goto/lib/cbmc/path/to/cbmc:/path/to/gnat2goto/gnat2goto/install
> ```
>
> Note for `make` users `/path/to/cbmc` will point to `lib/cbmc/src/cbmc`. For `cmake` it will be something like `lib/cbmc/build/bin`.

testsuite/gnat2goto/README.md
thk123: Personally find the `--diffs` a bit overwhelming for a first run, perhaps I could link here to fuller docs about the testuite if such a file exists?

chrisr-diffblue: Yeah, it's a bit overly verbose :-) Happy either way, I was mostly just checking you knew about those options.

thk123: Couldn't find any so mentioned it as an alternative to inspecting diffs. Personally I'd have `--diffs` only show difference in FAIL and UPASS (since XFAIL diffs are constant).

thk123:
@chrisr-diffblue @martin-cs  comments addressed, please confirm you're happy with my resolution to cbmc path question.

PR #330, opened 2020-01-22
Update tests with ordered properties

thk123:
Brings in diffblue/cbmc#5214

Previously propreties that differed only by name would be in a undefined order. Now they are sorted first by name, then by number. Expected output is updated accordingly.


thk123:
@chrisr-diffblue now pointing at `diffblue/cbmc:develop`

thk123:
@chrisr-diffblue @martin-cs I don't have write access. Not sure if I _should_, but this is ready to merge :)

chrisr-diffblue:
@thk123 Argh, apologies - I *thought* I'd added you to the appropriate group, but it turns out I hadn't actually clicked the extra "yes, I really do mean to do this" button. You should now have received an invite from Github.

PR #334, opened 2020-01-29
enable test for Non-literal bound #54

spanners:
Non-literal bound is now supported (i.e. 'First and 'Last), so this
patch creates a test to demonstrate this.

* tests/nonliteral_bound/
Add test.


spanners:
Hold on a moment - this does not catch Ekind `E_Signed_Integer_Subtype`. I will amend - until then please hold off from reviewing.

PR #327, opened 2020-01-14
Remove XFAIL test.opt files for passing tests #108

spanners:
* tests/attrib_size_object/test.out
* tests/attrib_size_object/test.opt
* tests/deferred_constant_use/test.opt
* tests/deferred_constant_use/test.out
* tests/enum_symbols_range_constraint_assignment/test.opt
* tests/enum_symbols_range_constraint_assignment/test.out
* tests/packages/test.opt
* tests/packages/test.out
* tests/record_extension/test.opt
* tests/record_extension/test.out
* tests/representation_clause_component_size_change/test.opt
* tests/representation_clause_component_size_change/test.out
* tests/size_attribute_clause/test.opt
* tests/size_attribute_clause/test.out
* tests/size_integer_and_objects/test.opt
* tests/size_integer_and_objects/test.out
Remove test.opt file and add the test.out output for each passing test

PR #335, opened 2020-01-30
Declare implicit array types

NlightNFotis:
This PR adds support for declaring Itypes (implicit types) for array declarations.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -59,14 +59,14 @@ Error message: Unsupported pragma: Check
 Nkind: N_Pragma
 --
 Occurs: 8 times
-Calling function: Process_Pragma_Declaration
-Error message: Unsupported pragma: Obsolescent
-Nkind: N_Pragma
---
-Occurs: 7 times
 Calling function: Do_Type_Reference
 Error message: Type of type not a type
 Nkind: N_Defining_Identifier
+--
+Occurs: 8 times
+Calling function: Process_Pragma_Declaration
+Error message: Unsupported pragma: Obsolescent
+Nkind: N_Pragma
 --
 Occurs: 6 times
 Calling function: Process_Declaration
@@ -108,6 +108,11 @@ Calling function: Do_Itype_Definition
 Error message: Unknown Ekind
 Nkind: N_Defining_Identifier
 --
+Occurs: 2 times
+Calling function: Do_Itype_Integer_Subtype
+Error message: Non-literal bound unsupported
+Nkind: N_Defining_Identifier
+--
 Occurs: 1 times
 Calling function: Do_Base_Range_Constraint
 Error message: unsupported lower range kind
@@ -132,11 +137,6 @@ Occurs: 1 times
 Calling function: Do_Function_Call
 Error message: function entity not defining identifier
 Nkind: N_Defining_Operator_Symbol
---
-Occurs: 1 times
-Calling function: Do_Itype_Integer_Subtype
-Error message: Non-literal bound unsupported
-Nkind: N_Defining_Identifier
 --
 Occurs: 1 times
 Calling function: Do_Op_Expon
@@ -1718,7 +1718,7 @@ Occurs: 10 times
 | GNU Ada (ada2goto) Assert_Failure sinfo.adb:495                          |
 Error detected at REDACTED
 --
-Occurs: 9 times
+Occurs: 8 times
 +===========================GNAT BUG DETECTED==============================+
 | GNU Ada (ada2goto) Constraint_Error Symbol_Table_Info.Symbol_Maps.Constant_Reference: key not in map|
 Error detected at REDACTED
--- a/gnat2goto/driver/gnat2goto_itypes.adb
+++ b/gnat2goto/driver/gnat2goto_itypes.adb
@@ -7,6 +7,7 @@ with Range_Check;           use Range_Check;
 with Symbol_Table_Info;     use Symbol_Table_Info;
 with Tree_Walk;             use Tree_Walk;
 with Follow;                use Follow;
+with Arrays;                use Arrays;
 
 package body Gnat2goto_Itypes is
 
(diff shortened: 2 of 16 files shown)


gnat2goto/driver/gnat2goto_itypes.adb
```diff
@@ -96,13 +98,35 @@ package body Gnat2goto_Itypes is
...
+         Object_Definition (Associated_Node_For_Itype (E));
+   begin
+      case Nkind (Var) is
+      when N_Constrained_Array_Definition =>
+         return Do_Constrained_Array_Definition (Var);
+      when N_Unconstrained_Array_Definition =>
+         return Do_Unconstrained_Array_Definition (Var);
+      when others =>
+         return Report_Unhandled_Node_Irep
+           (E, "Do_Itype_Array_Type", "Unknown array type");
```
chrisr-diffblue: Can you make this Report_Unhandled actually report *which* NKind is not handled? This is important for reviewing the list of unsupported features and making sure all are visible (otherwise different NKinds would get counted as one type)

hannes-steffenhagen-diffblue:
```suggestion
           (E, "Do_Itype_Array_Type", "Unknown array type " & Node_Kind'Image (Nkind (Var)));
```

NlightNFotis:
@chrisr-diffblue @hannes-steffenhagen-diffblue Can I get a re-review on this one please?

experiments/golden-results/StratoX-summary.txt
```diff
@@ -59,15 +59,15 @@ Error message: Unsupported pragma: Check
...
---
-Occurs: 7 times
 Calling function: Do_Type_Reference
 Error message: Type of type not a type
 Nkind: N_Defining_Identifier
 --
+Occurs: 8 times
+Calling function: Process_Pragma_Declaration
+Error message: Unsupported pragma: Obsolescent
+Nkind: N_Pragma
```
hannes-steffenhagen-diffblue: Most of the golden results changes appear to be just things moving around for some weird reason. Very few actual changes.

chrisr-diffblue: That looks like a move, but actually its because "Do_Type_Reference" went fomr 7 -> 8 times...

experiments/golden-results/StratoX-summary.txt
```diff
@@ -134,11 +139,6 @@ Error message: function entity not defining identifier
 Nkind: N_Defining_Operator_Symbol
 --
 Occurs: 1 times
-Calling function: Do_Itype_Integer_Subtype
```
chrisr-diffblue: Another one where we get an extra occurence

experiments/golden-results/StratoX-summary.txt
NlightNFotis: @chrisr-diffblue Yes, I believe this is because some of the nodes that we now handle were not handled before. This means that some nodes in these projects now propagate, and they might result into some previously unreported errors showing up.

gnat2goto/driver/tree_walk.adb
```diff
@@ -5118,6 +5118,7 @@ package body Tree_Walk is
         (if Ekind (E) = E_Access_Subtype then Etype (E) else E);
       Type_Id : constant Symbol_Id := Intern (Type_Name);
    begin
+      Declare_Itype (E);
```
hannes-steffenhagen-diffblue: This is the important bit; We now just proactively declare (possible) Itypes whenever we see them (`Declare_Itype` will take non-Itypes and just ignore them). This way, at least theoretically, we should always have types in the symbol table as soon as we reference them.

testsuite/gnat2goto/tests/function_pointer/test.out
```diff
@@ -1,2 +1,25 @@
+Standard_Output from gnat2goto test:
```
chrisr-diffblue: What's going on with these two tests? This looks like the tests are now producing missing feature reports which they were not before.... Were this test and procedure_pointer both passing before this change?

gnat2goto/driver/gnat2goto_itypes.adb
```diff
@@ -96,13 +98,36 @@ package body Gnat2goto_Itypes is
                                                     "Unknown Ekind"));
    end Do_Itype_Definition;
 
+   ----------------------------
+   -- Do_Itype_Array_Type --
+   ----------------------------
+
+   function Do_Itype_Array_Type (E : Entity_Id) return Irep is
+      Var : constant Node_Id :=
```
hannes-steffenhagen-diffblue: This should get a better name than `Var`. I don’t know, something like `Type_Of_Underlyign_Array` or something of the sort?

testsuite/gnat2goto/tests/function_pointer/test.out
hannes-steffenhagen-diffblue: It’s basically just Itypes that we never supported, but we just “happened” to pass these tests without missing feature reports or crashes because we don’t actually _use_ these types anywhere in the resulting goto program.

I.e. as a side effect of some of the changes we’ve made in this PR some things that didn’t work before now _visibly_ don’t work.

experiments/golden-results/StratoX-summary.txt
hannes-steffenhagen-diffblue: @chrisr-diffblue Keep in mind that the changes made in this PR cause gnat2goto to not crash in some circumstances where it would’ve crashed before.

testsuite/gnat2goto/tests/function_pointer/test.out
chrisr-diffblue: OK. There's recently been some debate on https://github.com/diffblue/gnat2goto/pull/327 about whether we should or should not add "missing feature" reports to expected outputs - I'm not sure we came to a conclusion. In this case, it's probably better to accept it in the expected output and keep the test passing, rather than turn a passing test into an xfail.

gnat2goto/driver/gnat2goto_itypes.adb
martin-cs: +10 what he said.

PR #337, opened 2020-04-19
Use enum pos

tjj2017:
Use enumeration literal position as their value rather than their representation value.  Using their position maintains an order relation between the literals of an enumeration whereas using their representation value does not necessarily.  If, in future, a representation value is needed (for low level I/O) a function could be provided to obtain the representation from an enumeration literals position.  Such a function exists in gnat but is possible a goto equivalent may be needed.


martin-cs:
@chrisr-diffblue : the Travis failure looks spurious, please can you kick it?

chrisr-diffblue:
I've just hit the restart button on that build.

martin-cs:
Thanks @chrisr-diffblue (and, of course @tjj2017 ).

PR #338, opened 2020-04-24
Translate withed units once only

tjj2017:
The branch name is a bit misleading.

The original purpose was to ensure a withed subprogram body is only processed once even if it is withed by multiple units.

However, as part of the update it soon became apparent that generic bodies were being translated even though they should not be (they are instantiated in the AST as non generic units).  This discovery further led to realising library level generic instantiations were not handled by gnat2goto.

All of these issues are intertwined so I have addressed them all in this branch.


martin-cs:
Again, CI failure on muen.  Weird.

martin-cs:
A tiny tweak to the golden results needed.  Rebasing and rerunning these should be sufficient.

PR #340, opened 2020-05-08
Iss 131 front end

tjj2017:
Modified front-end so that it does not abort with size too small message.

Commented out faulty attribute_size check pending improved algorithm.

This branch will be subsumed by branch iss_131, when the improved attribute_size algorithm is completed.  Unless the suppression of the size too small message is required before branch iss_131 is ready for merging, do not merge this branch.  Merge iss_131 instead.


tjj2017:
This PR has now been superseded by the PR #343

PR #342, opened 2020-05-19
Change the URLs for GNAT GPL 2016 and GNATCOL 2016 in .travis.yml

martin-cs:
The AdaCore website seems to have migrated to a new AWS S3 based
storage so the older URLs.  I have checked and files have the same
SHA1 as before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ dist: trusty
 compiler: gcc
 
 env:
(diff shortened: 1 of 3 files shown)


martin-cs:
I have added a fix to the testsuite expected results as another commit.  It fits with the general spirit of "fixing up CI".

PR #343, opened 2020-05-19
Iss 131 ready

tjj2017:
This PR supersedes PR #342 
It fixes Issue 131 and introduces a new model for attribute_size.


martin-cs:
Needs an actual rebase as the results of `iss_131_vanilla` don't match due to other things merged into master.

PR #353, opened 2020-06-15
Record aggregates

tjj2017:
This patch fixes two issues which cause gnat2goto to fail when processing the example pub/sub system.

They both related to record aggregates, particularly when the record is discriminated.
The first issue is fixed by using the component name string rather than its node id as the front-end seems to frequently rewrite such nodes giving them a different node id.
The other failure, due to an unsatisfied precondition is solved by relaxing the precondition to allow I_Code_Assert Ireps.

The patch does not solve all the problems, i.e. others => <> in a record still causes gnat2goto to fail a precondition and the symbol table generated seems to be very large and cbmc fails when run on the symbol table.  However, it does solve the immediate problem of the gnat2goto failures when processing the example pub/sub system.

### DIFF
--- a/gnat2goto/driver/goto_utils.ads
+++ b/gnat2goto/driver/goto_utils.ads
@@ -92,7 +92,7 @@ package GOTO_Utils is
                                        A_Symbol_Table : in out Symbol_Table)
                                        return Symbol
      with Pre => (Kind (Symbol_Type) = I_Code_Type
-                  and then (Kind (Value) = I_Code_Block
+                  and then (Kind (Value) in I_Code_Block | I_Code_Assert
                     or else Value = Ireps.Empty));
 
    function Create_Fun_Parameter (Fun_Name : String; Param_Name : String;
--- a/testsuite/gnat2goto/tests/record_aggregates_box/test.opt
+++ b/testsuite/gnat2goto/tests/record_aggregates_box/test.opt
@@ -1,0 +1,2 @@
+ALL XFAIL gnt2goto Failed precondition from tree_walk.ads
+
--- a/gnat2goto/driver/tree_walk.adb
+++ b/gnat2goto/driver/tree_walk.adb
@@ -687,8 +687,11 @@ package body Tree_Walk is
                                                   "Actual iter not present");
                      return Struct_Expr;
                   end if;
-                  if Defining_Identifier (Substruct_Component_List) /=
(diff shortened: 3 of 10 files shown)


gnat2goto/driver/goto_utils.ads
```diff
@@ -92,7 +92,7 @@ package GOTO_Utils is
                                        A_Symbol_Table : in out Symbol_Table)
                                        return Symbol
      with Pre => (Kind (Symbol_Type) = I_Code_Type
-                  and then (Kind (Value) = I_Code_Block
+                  and then (Kind (Value) in I_Code_Block | I_Code_Assert
```
martin-cs, line 95: Weakening preconditions always makes me a little nervous.  Why is does this solve the problem?

testsuite/gnat2goto/tests/record_aggregates_box/test.opt
```diff
@@ -0,0 +1,2 @@
+ALL XFAIL gnt2goto Failed precondition from tree_walk.ads
```
martin-cs, line 1: In CI this seems to pass...

gnat2goto/driver/goto_utils.ads
tjj2017: The precondition was for an I_Code_Block, An assignment range check function is created using the above function when the first assignment is encountered in a program (it could be an initialisation expression rather than an assignment statement.  Thereafter the same check function is used for all assignments that need range checking.  Normally, the kind of the value passed to this subprogram when the range check function is first created is an I_Code_Block, but if the first range check that is required is for an aggregate assignment, the kind of the value is an I_Code_Assert.

testsuite/gnat2goto/tests/record_aggregates_box/test.opt
tjj2017: I put the erroneous output in the test.out file by mistake.  Gnat2goto should not fail the precondition.  The test is legal Ada.  I have committed a test.out file which closer represents the expected output.

gnat2goto/driver/goto_utils.ads
martin-cs: OK. Fair.

testsuite/gnat2goto/tests/record_aggregates_box/test.opt
martin-cs: OK.

PR #358, opened 2020-09-02
Model size

tjj2017:
First stage of an implementation to record and publish (via attribute 'Size) the ASVAT model size of an object as opposed to its code size.

### DIFF
--- a/testsuite/gnat2goto/tests/iss_131_vanilla/test.out
+++ b/testsuite/gnat2goto/tests/iss_131_vanilla/test.out
@@ -1,108 +1,108 @@
 Standard_Output from gnat2goto iss_131:
 ----------At: Do_Attribute_Size----------
 ----------The size of the composite is not known by the front-end. Use a size representation clause on its declaration----------
-N_Identifier "r4" (Node_Id=3664) (source)
- Parent = N_Attribute_Reference (Node_Id=6769)
- Sloc = 17175  iss_131.adb:273:19
+N_Identifier "r4" (Node_Id=3674) (source)
+ Parent = N_Attribute_Reference (Node_Id=6799)
+ Sloc = 17335  iss_131.adb:276:19
  Chars = "r4" (Name_Id=300001470)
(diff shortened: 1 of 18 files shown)


testsuite/gnat2goto/tests/iss_131_vanilla/test.out
```diff
@@ -125,51 +125,53 @@ N_Identifier "vsf3_ns" (Node_Id=3880) (source,analyzed)
...
-[iss_131.assertion.47] line 329 assertion SFV3_No_Size_Rep'Size = 8: FAILURE
-[iss_131.assertion.48] line 330 assertion VST3_NS'Size = 64: FAILURE
-[iss_131.assertion.49] line 331 assertion VSF3_NS'Size = 32: FAILURE
-[iss_131.assertion.50] line 334 assertion A_T'Size = 8: SUCCESS
-[iss_131.assertion.51] line 335 assertion VA'Size = 8: SUCCESS
-[iss_131.assertion.52] line 336 assertion VA (1)'Size = 2: SUCCESS
-[iss_131.assertion.53] line 337 assertion VA (0 .. 2)'Size = 8: SUCCESS
-[iss_131.assertion.54] line 339 assertion Vna'Component_size = 1: SUCCESS
-[iss_131.assertion.55] line 340 assertion neigh_array'component_size = 1: SUCCESS
+[iss_131.assertion.9] line 252 assertion Vna (barbara)'size = 1: FAILURE
```
martin-cs, line 128: This seems to have changed status.  I am not sure if there are others.  Is this intended?

martin-cs (approved):

testsuite/gnat2goto/tests/iss_131_vanilla/test.out
tjj2017: Yes, there are some changes because we use the model size instead of the front-end derived size.

martin-cs: OK, as long as you are happy with these that is fine.

PR #359, opened 2020-12-07
Update the cbmc submodule to the current develop

martin-cs:
This is a test to see what (if anything) breaks.


martin-cs:
OK, so, a load of stuff breaks but CI does run.

martin-cs:
Replaced by https://github.com/diffblue/gnat2goto/pull/377

Issue #122, opened 2019-02-11
"Make_Assert_Call"  likely redundant


sonodtt:
It seems likely that the process of creating and then calling an assert function via "Make_Assert_Call" can be replaced with a 'make assert expression' via simple binding of an expression to an 
New_Irep (I_Code_Assert) via ' Set_Assertion(A_Irep1, Make_Op_Eq(...) )'

This would avoid generating 
-	the assert function 
-	the function call at call site

While the code saving would be minor - it would be more consistent with other locations e.g. " Do_Pragma_Assert_or_Assume" where assertions are generated simply via 
"Set_Assertion" as outlined above.   

Another concern is that there may be  cbmc warnings upon duplicated symbols in the symbol table when an asssert call is generated. This does not occur atm - just a potential issue along the lines of warnings we have seen before. 

The json template has this
{
  "parent": "code",
  "sub": [
    {"schema": "expr", "friendly_name": "assertion"}
  ],
  "namedSub": {
    "statement": {"constant": "assert", "type": "string"}
  }
}

It might be preferable to go via this interface rather than generating an assert function directly via use of "__CPROVER_assert" - this seems to have been the original design elsewhere in "tree_walk.adb"
 
• In long form:
Creating a simple c assert statement requires only something along these lines:

         Op_Equal_Var1 :=
           Make_Op_Eq (Rhs             => Make_Int_Constant_Expr("0"),
                       Lhs             => Var1_Symbol_Irep,
                       Source_Location => 0,
                       Overflow_Check  => False,
                       I_Type          => Make_Bool_Type,
                       Range_Check     => False);

         A_Irep1 := New_Irep (I_Code_Assert);
         Set_Assertion(A_Irep1, Op_Equal_Var1);
         Append_Op (Dummy_Body, A_Irep1);

This creates e.g.
" assert(var1_user == 0);"  in the  generated code. 

This was verified via 
https://github.com/sonodtt/gnat2goto/tree/irep_debug_framework_v0

PR #361, opened 2020-12-23
Issue 122

tjj2017:
The changes in this branch address the shortcomings raised in issue #122 to do with subprogram renaming.
The changes also address a related shortcoming of not accepting an operator as a function call.

The changes fix the the the problem posed by the example given in issue #122 as far as gnat2goto is concerned but, because no body is parsed for Interfaces.Shift_Left (an imported intrinsic function) cbmc raises a warning that no body is supplied for Shift_Left.

These changes also seem to produce a few extra unsupported feature reports from libsparkcrypto (the use of the attribute 'Loop_Entry and the use of the concatenation operator).  Neither of these have been or are currently supported by gnat2goto, so gnat2goto must be parsing more of the source code than before these changes.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -119,11 +119,6 @@ Error message: type not in symbol table after declaration
 Nkind: N_Full_Type_Declaration
 --
 Occurs: 1 times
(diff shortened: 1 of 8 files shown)


martin-cs (approved):
Looks good; thanks for this.

PR #365, opened 2021-01-18
Pragma initializes

jeremycrisp:
Added test case for Pragma Initializes and 
updated gnat2goto code to ignore the pragma rather than indicate that it is unsupported (with an explanation)


martin-cs:
CI fails because this removes some unsupported features from Muen.  See the end of the CI log for details.

testsuite/gnat2goto/tests/pragma_initializes/test.out
```diff
@@ -0,0 +1,8 @@
+[init.assertion.1] line 8 assertion NewVar1 = PragmaInitVars.Variable1: FAILURE
+[init.assertion.2] line 9 assertion NewVar2 = PragmaInitVars.Variable2: FAILURE
```
martin-cs, line 2: Good to see both failures and successes.

martin-cs (approved):
Find; can merge once CI passes.

martin-cs:
@jeremycrisp please can you add the diffs from the CI's `list_unsupported` so that we can merge this.

PR #366, opened 2021-01-26
added test case and updated code for pragma atomic

jeremycrisp:
Added a test for pragma Atomic and Atomic_Components.
updated gnat2goto to ignore the pragmas rather than report unsupported.


martin-cs (approved):
Looks good; thank you.

PR #367, opened 2021-01-26
Priv aggr/working

tjj2017:
This change addresses the problem described in Issue #132 from Do_Aggrgate_Literal.
It also resolves a problem described in Issue 52.


martin-cs (changes requested):
Code changes look good but somewhere along the line this seems to have picked up a load of commits that have already been merged.  Please can we resolve this?  Either by rebase or I am happy to work with whoever to unpick what has happened.

tjj2017:
I don't know what has happened. I had rebased to the current master.  I then went to push to my repository and git reported that I was behind the version on the repository which was only pushed up yesterday.  I thought perhaps Jeremy had done something on the branch so I rebased to the remote branch.  I think this is where the extra bits and pieces came in but I don't understand why.

tjj2017:
This PR has been replaced with PR #368 on a clean branch.

PR #52, opened 2018-10-25
Add statement handling exception fallback

hannes-steffenhagen-diffblue:
If an exception occured during exception handling we'd
halt compilation. With this commit, what we do instead is
report the error and continue on.

The result of the compilation will be ultimately not very useful,
but at least with this we can find out all the places where something
is going wrong rather easily.

For example, with the AdaYaml script we now get output like this:

```
Warning: Subprogram yaml__version_major not in symbol table
Warning: Subprogram yaml__version_minor not in symbol table
Warning: Subprogram yaml__version_patch not in symbol table
N_Use_Type_Clause (Node_Id=2295) (source,analyzed)
 Sloc = 8534  yaml.adb:12:4
 Subtype_Marks = List (List_Id=-99999978)
 Hidden_By_Use_Clause = <no elist>
 Used_Operations = (Elist_Id=100000158)
raised PROGRAM_ERROR : tree_walk.adb:3853 explicit raise
Call stack traceback locations:
0x960ee6 0x961099 0x958eaa 0x948532 0x903ad6 0x903846 0x10e49a4 0x10e62c6 0x405ef2 0x408635 0x7f7c941fdb95 0x403538 0xfffffffffffffffe

raised PROGRAM_ERROR : tree_walk.adb:1239 explicit raise
Call stack traceback locations:
0x94b199 0x9453cb 0x943e72 0x94ae0a 0x957a4c 0x960a8d 0x961099 0x94d2a3 0x960b48 0x958f53 0x958237 0x960d36 0x961099 0x958eaa 0x948532 0x903ad6 0x903846 0x10e49a4 0x10e62c6 0x405ef2 0x408635 0x7f7c941fdb95 0x403538 0xfffffffffffffffe

Warning: Subprogram yaml__default_properties not in symbol table
raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed precondition from goto_utils.ads:37
Call stack traceback locations:
0x1140fbf 0x6e4683 0x94c99b 0x94aa11 0x94b448 0x94b02a 0x957a4c 0x960a8d 0x961099 0x94d2a3 0x960b48 0x958f53 0x958237 0x960d36 0x961099 0x958eaa 0x948532 0x903ad6 0x903846 0x10e49a4 0x10e62c6 0x405ef2 0x408635 0x7f7c941fdb95 0x403538 0xfffffffffffffffe

Warning: Subprogram yaml__is_empty not in symbol table
Warning: Casestatements unhandled
[...]

PR #368, opened 2021-01-27
Priv aggr/final

tjj2017:
This PR replaces PR #367 with a different branch because of the inconsistent merges discussed on PR #367.

This change addresses the problem described in Issue #132 from Do_Aggrgate_Literal.
It also resolves a problem described in Issue #52.

PR #367 is no closed.


martin-cs (approved):
Thanks.  In future please can test cases have a few VCs that should fail / have counter-examples.

PR #369, opened 2021-01-28
Rec data/working

tjj2017:
The declaration of a recursive data type was causing infinite recursion and a stack overflow.
This change addresses this issue for simple recursive data type declarations.


martin-cs (approved):
Looks good; thanks.

PR #370, opened 2021-02-02
Fix up/final

tjj2017:
Handles private and incomplete declarations whose full type declaration uses type declared between the partial declaration and its completion.

This feature was required to address issue #194.


testsuite/gnat2goto/tests/loop_array_range/loop_array_range.adb
```diff
@@ -8,5 +8,5 @@ begin
       Sum := Sum + My_Arr(I);
    end loop;
 
-   pragma Assert(Sum=6);
+   pragma Assert(Sum=10);
```
martin-cs, line 11: Why change this?  Why not have both?  A test can still pass if CBMC says `VERIFICATION FAILED`.  It is better to have a mix of tests TBH.

testsuite/gnat2goto/tests/range_third/test.opt
```diff
@@ -0,0 +1,2 @@
+ALL XFAIL Doesn't crash anymore but cbmc does not perform range assertion checks 
+
```
martin-cs, line 2: @jeremycrisp worth keeping in mind.

martin-cs (approved):
A couple of questions about tests but basically, yes.

testsuite/gnat2goto/tests/range_third/test.opt
tjj2017: I felt the purpose of the test was to check the loop and therefore felt a positive result was required.  I have added a negative test in which the assertion fails.

martin-cs (approved):
Great; thank you.

tjj2017:
Don't merge yet I have just located some debug code I have left in.  I will push again.

PR #371, opened 2021-02-08
Iss 195

tjj2017:
This change addresses the problem addressed in Issue #195.

The front-end "Chars" function was being applied to representation clause nodes which have no Chars field causing an assertion failure at sin line 394.

The change effects the golden results significantly as many more units are now being processed.


martin-cs (approved):
Looks good; big change to the tests but I think it is a net gain as we cover a lot more code.

PR #372, opened 2021-02-08
Ada checks

jeremycrisp:
added test cases for some ada runtime checks and updated gnat2goto to add one function call per type of check


martin-cs (approved):
Please can you rebase the PR so that there isn't a merge commit in it?

gnat2goto/driver/range_check.adb
```diff
@@ -184,7 +184,8 @@ package body Range_Check is
       --  }
       function Build_Assert_Function return Symbol
       is
-         Func_Name : constant String := Fresh_Var_Name ("division_check");
+         Func_Name : constant String := "__division_check";
+         --  Fresh_Var_Name ("division_check");
```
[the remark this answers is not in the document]
jeremycrisp: commented out code removed

gnat2goto/driver/range_check.adb
```diff
@@ -352,8 +354,8 @@ package body Range_Check is
       --  }
       function Build_Assert_Function return Symbol
       is
-         Func_Name : constant String :=
-           Fresh_Var_Name ("range_check_" & File_Name_Without_Extension (N));
+         Func_Name : constant String := ("range_check" & Check_Name);
+         --  Fresh_Var_Name ("range_check_" & File_Name_Without_Extension (N));
```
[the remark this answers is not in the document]
jeremycrisp: commeneted out code removed

testsuite/gnat2goto/tests/ada_rtc_division/test.out
```diff
@@ -0,0 +1,8 @@
+[assertion.1] file division_check.adb line 15 Ada Check assertion: SUCCESS
+[division_check.assertion.1] line 15 Ada Check assertion: FAILURE
+[division_check.assertion.6] line 15 Ada Check assertion: SUCCESS
```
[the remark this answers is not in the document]
martin-cs: As per discussion -- changing this would be good but should be a separate PR.

martin-cs (approved):
Approve but:

1. Please can you fix the location information in the generated `__CPROVER_Ada_...` functions.
2. Please can you wait merging this until I have tracked down what is going on with the `subprog_renaming` tests.

martin-cs:
3. Please can you fix up the line number changes in the golden results -- sorry.

martin-cs (changes requested):
The issue with `subprog_renaming` test is cause by `range_check__CPROVER_Ada_Overflow_Check` having parameters that are `short int`s but being called with `long int` parameters.  These are then truncated so that `-2147483648` becomes `0` and `2147483647` becomes `-1` resulting an `assert(false); assume(false);`.

Also please can the range check be called `__CPROVER_Ada_range_check_...` rather than the other way around.

martin-cs (commented):
Please can you have a look at the tests `range_second` and `range_third`.  Both are `XFAIL` at the moment but I would hope this patch will fix that.

martin-cs (changes requested):
One last change : all of the functions we generate internally must have `__CPROVER_Ada` at the start of them.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -73,31 +73,26 @@ Calling function: Do_Expression
 Error message: Index of string unsupported
 Nkind: N_Indexed_Component
 --
-Occurs: 30 times
```
martin-cs, line 76: Why are there now fewer of these?

testsuite/gnat2goto/tests/nat_and_pos/nat_and_pos.adb
```diff
@@ -8,16 +8,14 @@ procedure Nat_And_Pos is
      with Annotate => (ASVAT, In_Type_Function),
      Import;
 begin
-   Var := -1;
-   pragma Assert (In_Type_Natural (Var));
```
martin-cs, line 12: Why remove these assertions?

martin-cs (approved):
Approve subject to reasonable answers to the questions.

testsuite/gnat2goto/tests/nat_and_pos/nat_and_pos.adb
jeremycrisp: The test has been reordered so that only the last test fails. the test of In_Type_Natural (-1) has been moved to the end of the tests.

martin-cs: Ah, yes, if they are checks on constants that makes sense.

testsuite/gnat2goto/tests/nat_and_pos/nat_and_pos.adb
```diff
@@ -8,16 +8,14 @@ procedure Nat_And_Pos is
...
    pragma Assert (In_Type_Natural (Var));
-   pragma Assert (In_Type_Positive (Var));
    Var := 1;
    pragma Assert (In_Type_Natural (Var));
    pragma Assert (In_Type_Positive (Var));
    Var := Integer'Last;
    pragma Assert (In_Type_Natural (Var));
    pragma Assert (In_Type_Positive (Var));
+   Var := -1;
+   pragma Assert (In_Type_Natural (Var));  -- This will cause a range check failure
```
martin-cs, line 20: It should but in the results it doesn't because the failure on line 12 blocks the rest of the execution.  My suggested fix would be to leave the assertions in the original order but

1. Create a local Boolean variable for each one and leave it uninitialised.
2. For each assertion, make it:
```
if Boolean_n then
  pragma Assert (...);
end if;
```

experiments/golden-results/Tokeneer-summary.txt
```diff
@@ -1225,17 +1225,17 @@ Error detected at REDACTED
 --
 Occurs: 1 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure failed precondition from tree_walk.ads:108|
+| GNU Ada (ada2goto) Assert_Failure range_check.adb:509                    |
```
martin-cs, line 1228: These look like crashes in the introduced code, can you easily see why this is happening?

experiments/golden-results/muen-summary.txt
```diff
@@ -206,6 +196,11 @@ Nkind: N_Pragma
 Occurs: 3 times
 Calling function: Process_Statement
 Error message: Unknown expression kind
+Nkind: N_Object_Renaming_Declaration
```
martin-cs, line 199: It's still not clear to me why the number of these decreases.

martin-cs (approved):
More tweak requests I'm afraid.

experiments/golden-results/Tokeneer-summary.txt
jeremycrisp: The issue is with system.ads, It was failing in the call to Do_Expression because the node kind was not N_Subexpr.
Now it is trying to do a range check first and the value of the range checked value is not a Class_Expr.

testsuite/gnat2goto/tests/nat_and_pos/nat_and_pos.adb
jeremycrisp: assertions have been placed in original order and if statements added as described.

testsuite/gnat2goto/tests/ada_rtc_overflow/overflow_check.adb
```diff
@@ -0,0 +1,26 @@
...
+   
+   --  test in range will successed
+   var4 := var1 + var2;
+
+   pragma Assert (var4 = -2_147_483_638);
+ 
+   -- expect failure check
+   pragma Assert (false);
+
+   -- test out of range will fail
```
martin-cs, line 23: Move this check above the `Assert(false)` please.

testsuite/gnat2goto/tests/ada_rtc_division/division_check_float.adb
```diff
@@ -0,0 +1,30 @@
...
+
+      Fl1 : float1;
+      Fl2 : float1;
+
+   begin
+
+      Fl1 := 2.0;
+      Fl2 := 2.0;
+
+      Fl1 := float1 (val1) / float1 (Val2);
```
martin-cs, line 15: I don't get why there is one failure and one success here?

testsuite/gnat2goto/tests/ada_rtc_division/division_check_float.adb
```diff
@@ -0,0 +1,30 @@
+package body Division_Check_Float is
+
+   procedure Float_Check (val1 : float; Val2 : float) is
```
martin-cs, line 3: Pedantry but could it be `Val1` rather than `val1`?

testsuite/gnat2goto/tests/ada_rtc_division/division_check_float.adb
```diff
@@ -0,0 +1,30 @@
...
+
+      Fl1 := 2.0;
+      Fl2 := 2.0;
+
+      Fl1 := float1 (val1) / float1 (Val2);
+
+      --  valid float division
+      pragma Assert (Fl1 = 1.0);
+
+      Fl2 := float1 (val1) - float1 (Val2);
```
martin-cs, line 20: Presumably this generates checks but they are trivial so CBMC omits them?

testsuite/gnat2goto/tests/ada_rtc_division/division_check_float.adb
```diff
@@ -0,0 +1,30 @@
...
+
+      --  add a failure to check
+      pragma Assert (false);
+
+      --  this will cause a Constraint Error
+      Fl1 := Fl1 / Fl2;
```
martin-cs, line 26: Why is this not in the results?

testsuite/gnat2goto/tests/ada_rtc_division/division_check_int.adb
```diff
@@ -0,0 +1,38 @@
...
+
+      pragma Assert (Int1 = 1);
+
+      Int2 := Int2 - (Int1 *2);
+
+      -- add an assert fail for check
+      pragma Assert (False);
+
+      --  this will cause a Constraint Error
+      --  Int1 := Int1 / Int2;
```
martin-cs, line 35: Rather than commented out this could be in an `if Nondet_Boolean then ...` block.

experiments/golden-results/StratoX-summary.txt
```diff
@@ -2110,7 +2115,7 @@ Error detected at REDACTED
 --
 Occurs: 1 times
 +===========================GNAT BUG DETECTED==============================+
-| GNU Ada (ada2goto) Assert_Failure failed precondition from range_check.ads:55|
+| GNU Ada (ada2goto) Assert_Failure failed precondition from range_check.ads:56|
```
martin-cs, line 2118: Please could you look at this, I am 99% sure it is a missing case in an precondition.  It is fine to add it and then throw an unsupported feature in the implementation of `Make_Range_Assert_Expr` if it is a lot of effort to support.

experiments/golden-results/Tokeneer-summary.txt
```diff
@@ -1275,17 +1280,17 @@ raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : failed precondition from arrays.ads:51
 --
 Occurs: 10 times
 <========================>
-raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : range_check.adb:467
+raised SYSTEM.ASSERTIONS.ASSERT_FAILURE : range_check.adb:509
```
martin-cs, line 1283: This looks like a similar issue with a precondition that is not quite right.

martin-cs (changes requested):
The golden results look good.  I like the test cases.  Sorry but some more changes would be good:

- There are some precondition failures in the golden results related to range checking.  They look quite simple, would you mind having a look?
- Between us I think we are going to need to find a way to make sure that all of the VCs are reported in the tests because there are a bunch that I think are missing but are actually probably just quietly dropped by CBMC.  Let's talk about this.

testsuite/gnat2goto/tests/ada_rtc_overflow/overflow_check.adb
```diff
@@ -0,0 +1,26 @@
...
+   var1 : Integer := Integer'First;
+   var2 : Integer := 10;
+   var3 : Integer := Integer'Last;
+   var4 : Integer;
+   var5 : Integer;
+   var6 : Integer;
+
+begin
+
+   --  test undetermined will succeed
```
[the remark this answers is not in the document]
martin-cs: Ignore; sorry!

testsuite/gnat2goto/tests/ada_rtc_division/division_check_int.adb
jeremycrisp: test updated as described

martin-cs (approved):
Have discussed this with @jeremycrisp and any further improvements will go in another PR as they will change the format (but not the content) of a number of other tests.

PR #373, opened 2021-02-15
fix gnat crash

jeremycrisp:
code fix for gnat crash in range_check.ads pre condition


martin-cs (approved):
Thought it might be something simple like that.  Thanks.

PR #376, opened 2021-03-05
Resolve an assertion failure cause by array initialisation without 'o…

martin-cs:
…thers'

The assertion made an incorrect assumption about the case when there are
elements assigned but no others statement.


tjj2017 (approved):
I agree these are the changes we made.

PR #380, opened 2021-03-25
Reduce the size of JSON files by printing in compact mode

martin-cs:
If you need the verbose mode, try running:

$ jq . whatever.json_symtab


jeremycrisp (approved):
I'm happy with the changes. The uncompacted file can be recovered with a simple comment.
Note the heap exhausted error seems to be generated randomly, I've seen it come and go in other golden results

PR #382, opened 2021-03-31
Unchecked conversion

jeremycrisp:
Added functionality for unchecked conversion and for 'valid attribute.

### DIFF
--- a/experiments/golden-results/StratoX-summary.txt
+++ b/experiments/golden-results/StratoX-summary.txt
@@ -13,6 +13,11 @@ Calling function: Process_Pragma_Declaration
 Error message: Unsupported pragma: Precondition
 Nkind: N_Pragma
 --
+Occurs: 444 times
+Calling function: Get_Type_Size
+Error message: Type has no valid size
+Nkind: N_Defining_Identifier
+--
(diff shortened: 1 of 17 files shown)


tjj2017 (approved):
ASVAT_Modelling.adb
Lines 509-534 The function Get_Type_Size is commented out and replaced by the same named function in lines 535-562.
The commented out version uses the ASVAT.Size_Model to obtain the size of entities whereas the active version uses RM_Size and Esize.  RM_Size (I think is the wrong value to use) and Esize do not work reliably and will often return 0.  They require the backend to fill in the detail.  Gnat2goto is our backend and does not back fill information into the Atree.  Hence, the introduction of the ASVAT.Size_Model.  Going forward the ASVAT.Size_Model should definitely be used rather than RM_Size and Esize, even if more errors are reported at the moment.

Lines 715-719 & 807-852 Bool_Types, pointer_types, string_types are not supported.  This is apparent from the messages from CI tests.

I will raise issues for the comments I have made but I will approve the current PR for the next release.

martin-cs:
Replaced by https://github.com/diffblue/gnat2goto/pull/401

PR #385, opened 2021-04-12
Unchecked conversion

martin-cs:
This is the same code as https://github.com/diffblue/gnat2goto/pull/382 . I'm picking this up after @jeremycrisp has moved on to other things.


martin-cs:
Replaced by https://github.com/diffblue/gnat2goto/pull/401

PR #392, opened 2021-07-22
Fix/runtime check polymorphism

martin-cs:
This avoids problems in linking where different files have checks with different signatures but the same name.  We just add the types into the name.


tjj2017 (approved):
All looks sensible.

PR #393, opened 2021-07-27
Rebased new array functionality with current master.

tjj2017:
Initial version of new array functionality.

18 regression tests currently failing.


martin-cs (approved):
It is not possible to review this in full detail but it is definitely a net improvement in many directions, so I am in favour of merging it.

PR #396, opened 2021-08-02
Add the name of the function to the variables generated in __CPROVER_…

martin-cs:
…start

Previously the variable for the return value was just called return' which
causes linking failure if two or more of the linked components had
autogenerated __CPROVER_start for functions with different return types.


tjj2017 (approved):
All looks fine

PR #400, opened 2021-08-21
This change addresses issue 314.

tjj2017:
It removes the bottleneck of creating a large string containing
the text version of the JSON Symbol before outputting the string
to the json_symtab file.

A JASON_Value Write procedure has been created in a child package
of GNATCOLL.JASON (GNATCOLL.JSON.File) which outputs the text version
of a JSON_Value directly to a given text file.

Further efficiencies may be possible by disposing of unwanted fields of the symbol table during
conversion of the symbol table to JSON rather than generating a new symbol table without
the unwanted fields.  If this necessary, I think this should be handled as new issue.

Similarly, but more difficult, one could consider outputting the JSON format directly from
the symbol table without using GNATCOLL.JSON.
This could avoid building a large in memory JSON structure representing the symbol table.
Again, think this is another issue.

### DIFF
--- a/gnat2goto/driver/driver.adb
+++ b/gnat2goto/driver/driver.adb
@@ -39,6 +39,7 @@ with Arrays;                use Arrays;
 with Gather_Irep_Symbols;
 
 with GNATCOLL.JSON;         use GNATCOLL.JSON;
+with GNATCOLL.JSON.File;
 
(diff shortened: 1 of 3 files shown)


martin-cs (approved):
Great; thank you.

PR #403, opened 2021-08-25
A new-location based way of generating names for temporary variables

martin-cs:
The previous system had an issue with including headers in different orders
leading to inconsistent numbering between different files.  By using the
location, symbols should be identical across different inclusion orders.


tjj2017 (approved):
The changes look fine.

I think they may be a further instance where there is a possible name conflict  associated with record declarations which uses a different mechanism.  I will raise an issue regarding the possible name conflict.

PR #413, opened 2021-08-31
Ireps 1571 1

tjj2017:
This change fixes at least one case of failed precondition at ireps.adb:1571


martin-cs (approved):
Good but please can you remove the changes to lib/cbmc ?

